PR #14, opened 2018-11-28
started adding cli utils to solo-kit

EItanya:


pkg/utils/protoutils/marshal.go
```diff
@@ -78,6 +78,16 @@ func UnmarshalMap(m map[string]interface{}, into proto.Message) error {
 	return UnmarshalBytes(data, into)
 }
 
+func MarshalJSON(from proto.Message) ([]byte, error) {
```
mitchdraft: this is a duplicate of MarshalBytes

pkg/utils/protoutils/marshal.go
```diff
@@ -78,6 +78,16 @@ func UnmarshalMap(m map[string]interface{}, into proto.Message) error {
...
+	buf := &bytes.Buffer{}
+	err := jsonpbMarshaler.Marshal(buf, from)
+	return buf.Bytes(), err
+}
+
+func UnmarshalJSON(data []byte, into proto.Message) error {
```
mitchdraft: duplicate of UnmarshalBytes

pkg/utils/cliutils/printer.go
```diff
@@ -0,0 +1,124 @@
...
+	b, err := yaml.JSONToYAML(jsn)
+	if err != nil {
+		return errors.Wrap(err, "unable to convert to YAML")
+	}
+	_, err = fmt.Fprintln(w, string(b))
+	return err
+}
+
+// PrintJSONList - prints the given list to io.Writer in JSON
+func PrintJSONList(data interface{}, w io.Writer) error {
```
mitchdraft: I think the marshal/unmarshal utils handle lists - is this needed?

pkg/utils/cliutils/printer.go
```diff
@@ -0,0 +1,124 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	_, err = fmt.Fprintln(w, "]")
+	return err
+}
+
+// PrintYAMLList - prints the given list to io.Writer in YAML
+func PrintYAMLList(data interface{}, w io.Writer) error {
```
mitchdraft: as above, we should not need to give yaml lists special treatment - it's still yaml

EItanya: True, but I think they have slightly disparate purposes. One is meant for an arbitary list while the other is meant specifically for a proto message, the list utils could probably go in a different file if anything, but I think they could have some use.

If we massage the data at all then we would most likely want to call the list functions afterwards as opposed to the proto functions.

PR #21, opened 2018-12-17
Solo kit cli

EItanya:


cmd/cli/flags/config.go
```diff
@@ -0,0 +1,30 @@
...
+const (
+	input          = "input"
+	input_default  = "api/v1/"
+	output         = "output"
+	output_default = "pkg/api/v1"
+	docs           = "docs"
+	docs_default   = "doc/docs"
+	config         = "config"
+	gogo           = "gogo"
+	gogo_default   = true
```
ilackarms: consider give the reader an idea of what these constants are going to be used for
e.g.:
```
inputPrefix          = "input" // check if a file has this string as a prefix
```

cmd/cli/util/config.go
```diff
@@ -0,0 +1,63 @@
...
+	"path"
+	"path/filepath"
+
+	"github.com/ghodss/yaml"
+	"github.com/solo-io/solo-kit/cmd/cli/options"
+	"github.com/solo-io/solo-kit/pkg/errors"
+)
+
+const (
+	SOLO_KIT_YAML = "solo-kit.yaml"
```
ilackarms: nit: `SoloKitConfigFilename` or something like that

cmd/cli/util/config.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func EnsureConfigFile(opts *options.Options) error {
+	if opts.ConfigFile == "" {
+		// Check the current file tree
+
+		searchPath := os.ExpandEnv("$GOPATH")
```
ilackarms: nit:  i think convention here is to use `os.Getenv("GOPATH")`

PR #22, opened 2018-12-17
wrap protoc rather than have protoc wrap us

ilackarms:
solo kit code generation can now be done directly from its own main file, no more `generate.sh`!

tested working with solo-projects and gloo
untested against supergloo

solves the problem of
- not being able to debug solo kit
- `generate.sh` technical debt

all original plugin code is preserved (this should not break current use of `protoc-gen-solo-kit`)


cmd/solo-kit-gen/main.go
```diff
@@ -0,0 +1,192 @@
...
+	"-I" + Gopath() + "/src/github.com/solo-io/solo-kit/api/external",
+}
+
+func main() {
+	if err := Main(); err != nil {
+		log.Fatalf("%v", err)
+	}
+}
+
+//
```
rickducott: remove me

pkg/cli/compile.go
```diff
@@ -0,0 +1,9 @@
+package cli
+
+import (
+	"github.com/golang/protobuf/protoc-gen-go/plugin"
+)
+
+func CreateRequest(dir string) (*plugin_go.CodeGeneratorRequest, error) {
```
rickducott: fix me

pkg/cli/compile_test.go
```diff
@@ -0,0 +1,20 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "github.com/solo-io/solo-kit/pkg/cli"
+)
+
+var _ = Describe("Compile", func() {
+	Describe("CreateRequest(dir)", func() {
+		It("invokes protoc to return the code generation request", func() {
+			dir := "/home/ilackarms/go/src/" +
```
rickducott: ilackarms

cmd/solo-kit-gen/main.go
mitchdraft: stray code here

cmd/solo-kit-gen/main.go
```diff
@@ -0,0 +1,192 @@
...
+
+	absoluteRoot, err := filepath.Abs(*relativeRoot)
+	if err != nil {
+		return err
+	}
+
+	var projectDirs []string
+
+	// discover all project.json
+	if err := filepath.Walk(absoluteRoot, func(path string, info os.FileInfo, err error) error {
```
mitchdraft: this seems fine for now, but if we ever add javascript code to one of these projects it will get false positives since most javascript projects include `projects.json`
Instead, I think this would be better as a list or a new convention, like "solokitproject.json"

cmd/solo-kit-gen/main.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/solo-io/solo-kit/pkg/errors"
+	"github.com/solo-io/solo-kit/pkg/utils/log"
+)
+
+func Gopath() string {
+	return os.Getenv("GOPATH")
+}
+
+var commonImports = []string{
+	"-I" + Gopath() + "/src",
```
rickducott: I thought the parameter syntax was "-I=...", so it seems like Gopath() is expected to be "=...", which would be weird. Is this missing a "="?

pkg/code-generator/cmd/plugin.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/solo-io/solo-kit/pkg/code-generator/codegen"
+	"github.com/solo-io/solo-kit/pkg/code-generator/docgen"
+	"github.com/solo-io/solo-kit/pkg/code-generator/model"
+	"github.com/solo-io/solo-kit/pkg/code-generator/parser"
+	"github.com/solo-io/solo-kit/pkg/errors"
+	"github.com/solo-io/solo-kit/pkg/utils/log"
+)
+
+const (
+	ParameterPrefix_OUTPUT   = "OUTPUT="
```
rickducott: nit: inconsistent style OUTPUT vs Generate

cmd/solo-kit-gen/main.go
EItanya: solo-kit.json?

cmd/solo-kit-gen/main.go
ilackarms: the `=` is optional

cmd/solo-kit-gen/main.go
ilackarms: agreed

PR #23, opened 2018-12-18
fix issues with doc generation

ilackarms:


cmd/solo-kit-gen/main.go
```diff
@@ -142,16 +155,119 @@ func run() error {
 	return nil
 }
 
+func GopathSrc() string {
+	return filepath.Join(os.Getenv("GOPATH"), "src")
+}
+var commonImports = []string{
+	GopathSrc(),
+	GopathSrc() + "/github.com/solo-io/solo-kit/api/external",
```
rickducott: nit: filepath join these

cmd/solo-kit-gen/main.go
```diff
@@ -57,9 +51,6 @@ func run() error {
 	}
 
 	for _, inDir := range projectDirs {
-		outDir := strings.Replace(inDir, "api", "pkg/api", -1)
-
-		imports := append(commonImports, "-I"+inDir)
 
 		tmpFile, err := ioutil.TempFile("", "solo-kit-gen-")
 		if err != nil {
```
rickducott: I think something is messed up with the spacing below this line (not caused by this PR)

rickducott: Nvm, just looks weird in github

cmd/solo-kit-gen/main.go
```diff
@@ -72,11 +63,31 @@ func run() error {
 
 		var descriptors []*descriptor.FileDescriptorSet
 
-		if err := filepath.Walk(inDir, func(path string, info os.FileInfo, err error) error {
-			if !strings.HasSuffix(path, ".proto") {
+		var projectGoPackage string
+		if err := filepath.Walk(inDir, func(protoFile string, info os.FileInfo, err error) error {
```
rickducott: This is a 30 line nested lambda, I think we should factor it out -- pretty confusing to figure out what this is doing.

PR #28, opened 2018-12-24
Docs support enums

ilackarms:
add support for enums to the proto doc generation in solo-kit


ilackarms:
this also handles nesting of messages and enums within messages

PR #39, opened 2018-12-27
Update template to use build tag

rickducott:
Testing with gloo, when upgrading to this revision and regenerating these tests get ignored by CI (https://github.com/solo-io/gloo/pull/191/files)


ilackarms:
do you want to apply the same change to the other test templates?

rickducott:
These are the ones that were slowing down the builds, we could add them elsewhere but going to just merge this in for now.

PR #40, opened 2018-12-27
Incorporate kubecontroller into solo-kit

marcogschmidt:
Moved the code from [kubecontroller repo](https://github.com/solo-io/kubecontroller) into solo-kit.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,14 @@
 
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


ilackarms:
simply adding a note here since this PR is on the subject of kubernetes caches:
we are winding up sharing the kubecontroller and kube resource client cache with the configmap, secret, endpoint, service, pod, etc clients, some of which use solo kit clients and some of which go directly to the kubernetes api (service watch, pod watch, ingress watch)

PR #42, opened 2018-12-28
generate hashing function for resources which allows omitting fields

ilackarms:
with this PR:

in the case your project name is `gloo.solo.io` and you wish to make an event loop (resource group) `gloo.solo.io` i.e. we now support un-prefixed package name as the name of a resource group


pkg/code-generator/codegen/generator.go
```diff
@@ -36,7 +36,7 @@ func GenerateFiles(project *model.Project, skipOutOfPackageFiles bool) (code_gen
 		files = append(files, fs...)
 	}
 	for _, grp := range project.ResourceGroups {
-		if skipOutOfPackageFiles && !strings.HasSuffix(grp.Name, "."+project.GroupName) {
+		if skipOutOfPackageFiles && !(strings.HasSuffix(grp.Name, "."+project.GroupName) || grp.Name == project.GroupName) {
```
rickducott: I expected this to become `!strings.HasSuffix(grp.Name, project.GroupName)` 

Works fine according to https://golang.org/pkg/strings/#HasSuffix

ilackarms: this is problematic if project group names themselves have overlaps. e.g. `supergloo.solo.io` and `gloo.solo.io`
that's why we need the prefix case for `.gloo.solo.io` as well as `==` case for `gloo.solo.io`

ilackarms:
the scope of this pr has grown beyond the original intent; i'm going to close this in favor of opening a few smaller ones

ilackarms:
reopening, will rebase this on #45 and #46 after they are merged.

the purpose of this PR is to enable ignoring certain fields when we hash objects. this allows us to ignore a field which the controller is expected to update (such as status) using proto options which allow us to generate customized `Hash()` functions for each resource

Gopkg.toml
```diff
@@ -48,6 +48,11 @@
   version = "1.2.0"
   name = "github.com/golang/protobuf"
 
+[[constraint]]
+  branch = "switch-to-gogo"
+  source = "https://github.com/ilackarms/protokit"
+  name = "github.com/pseudomuto/protokit"
```
ilackarms: the reason for this change is that `github.com/pseudomuto/protokit` works entirely off of `golang/protobuf`.

however, all of our protos are based off of `gogo/protobuf`, which creates a compile error when trying to use our generated code `solo-kit.pb.go` with `protokit` (which we rely on for convenience functions to parse the proto code gen request).

this switches solo-kit to use a fork of protokit where `golang` is replaced with `gogo`

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -39,6 +40,22 @@ func (r *{{ .Name }}) SetMetadata(meta core.Metadata) {
...
+{{- range .Fields }}
+	{{- if not ( or (eq .Name "metadata") (eq .Name "status") .IsOneof .SkipHashing ) }}
+		r.{{ upper_camel .Name }},
+	{{- end }}
+{{- end}}
+{{- range .Oneofs }}
+		r.{{ upper_camel .Name }},
+{{- end}}
+	)
+}
```
ilackarms: here we are now generating a hashing function for every resource
this allows us to simplify the code in `Snapshot` as well as omit fields from hashing specified by the user (as a custom proto field option)

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -103,6 +120,20 @@ func (list {{ .Name }}List) Clone() {{ .Name }}List {
...
+
+func (list {{ .Name }}List) AsInterfaces() []interface{}{
+	var asInterfaces []interface{}
+	list.Each(func(element *{{ .Name }}) {
+		asInterfaces = append(asInterfaces, element)
+	})
+	return asInterfaces
+}
+
```
ilackarms: convenience functions

pkg/code-generator/model/project.go
```diff
@@ -65,10 +67,15 @@ type Resource struct {
...
-	Original *descriptor.FieldDescriptorProto
+	Name        string
+	TypeName    string
+	IsOneof     bool // we ignore oneof fields in test generation
+	SkipHashing bool // skip this field when hashing the resource
+	Original    *descriptor.FieldDescriptorProto
+}
+
+type Oneof struct {
+	Name string
```
ilackarms: because we are hashing individual fields rather than the whole object, we must account for `Oneof` fields

pkg/utils/hashutils/hash.go
```diff
@@ -0,0 +1,32 @@
...
+	var hashes []uint64
+	for _, v := range values {
+		hashes = append(hashes, hashValue(v))
+	}
+	return hashValue(hashes)
+}
+
+func hashValue(val interface{}) uint64 {
+	if hasher, ok := val.(Hasher); ok {
+		return hasher.Hash()
```
ilackarms: this is how we utilize the new custom generated hash functions

Gopkg.toml
rickducott: how does versioning work on this fork (i don't see a version constraint here)

yuval-k: also, this doesn't seem to be temporary. why not use your fork as the dependency?

ilackarms: i'll make it a permanent dependency and update this pr

ilackarms: nevermind, update in next pr

PR #47, opened 2018-12-31
map types.Any proto

ilackarms:
Signed-off-by: Scott Weiss <sdw35@cornell.edu>


ilackarms:
this is to enable using google's `protobuf.Any` type with `gogo`

PR #49, opened 2018-12-31
Namespace scoped resource clients and watch improvements

marcogschmidt:
This PR includes two main (non-breaking) changes to the Kubernetes resource clients:

**1 - Resource clients can now be namespace-scoped**
The `KubeResourceClientFactory` can now be configured with a namespace whitelist. When the client is `Register`ed, it will create a `SharedIndexedInformer` for each of the namespaces in this list (more specifically, each of the informer `ListWatch`es will target only one namespace). Previously, only one informer was created and it would list/watch all namespaces. The problem with this was that it required the client to have cluster-wide access to resources. Now we can build resource clients for users that have access to only a limited number of namespaces. E.g. if we have a user that can only access `ChangeSets` in the `gloo-system` namespace we can do:
```go
csClient, err := v1.NewChangeSetClientWithToken(&factory.KubeResourceClientFactory{
		...
		NamespaceWhitelist:      []string{"gloo-system"},
	}, <token_that_only_has_access_to_changesets_in_gloo-system>)
```

**2 - Resource client watches will only receive events for the relevant resource/namespace**
A new `ResourceEventHandler` now provides more context as to which resource was actually updates. The `ResourceClient` uses the additional information to filter out events that are not relevant to the specific watch. Earlier any event on the shared cache would notify all the watches created by clients that shared that cache. Now e.g. a watch created for the `csClient` in the example above would be notified only when an event occurred on some `ChangeSet` in the `gloo-system` namespace.

**Other minor changes**
- a `KubeResourceClientFactory` can now specify its own `resyncPeriod` for informer cache refreshes
- added a 10s timeout when calling the `KubeController.Run(...)` method in order to avoid blocking the whole process if the initial informer cache synchronization cannot be performed (e.g. if the client does not have access to a resource)

**NOTE:** All changes are backwards compatible:
- if no `namespaces` list is specified on the `KubeResourceClientFactory`, we default to all namespaces
- if no `resyncPeriod` is specified on the `KubeResourceClientFactory`, we default to the `SharedInformerFactory` resync period (12h, as before)

### DIFF
--- a/pkg/api/v1/clients/factory/resource_client_factory.go
+++ b/pkg/api/v1/clients/factory/resource_client_factory.go
@@ -63,10 +63,11 @@ func newResourceClient(factory ResourceClientFactory, params NewResourceClientPa
 		// Validate namespace list:
 		// 1. If no namespace list was provided, default to all namespaces
 		// 2. Error if namespace list contains the empty string plus other values
-		namespaces := opts.Namespaces
-		if namespaces == nil || len(namespaces) == 0 {
+		namespaces := opts.NamespaceWhitelist
+		if len(namespaces) == 0 {
 			namespaces = []string{metaV1.NamespaceAll}
-		} else if stringutils.ContainsString(metaV1.NamespaceAll, namespaces) && len(namespaces) > 1 {
+		}
+		if len(namespaces) > 1 && stringutils.ContainsString(metaV1.NamespaceAll, namespaces) {
 			return nil, fmt.Errorf("the kube resource client namespace list must contain either "+
 				"the empty string (all namespaces) or multiple non-empty strings. Found both: %v", namespaces)
 		}
@@ -102,12 +103,12 @@ type ResourceClientFactory interface {
 // Clients built with this factory will be able to access only resources the given namespace list. If no value is provided,
 // clients will be able to access resources in all namespaces.
 type KubeResourceClientFactory struct {
-	Crd             crd.Crd
-	Cfg             *rest.Config
-	SharedCache     *kube.KubeCache
-	SkipCrdCreation bool
-	Namespaces      []string
-	ResyncPeriod    time.Duration
+	Crd                crd.Crd
+	Cfg                *rest.Config
+	SharedCache        *kube.KubeCache
+	SkipCrdCreation    bool
+	NamespaceWhitelist []string
+	ResyncPeriod       time.Duration
 }
 
 func (f *KubeResourceClientFactory) NewResourceClient(params NewResourceClientParams) (clients.ResourceClient, error) {
--- a/pkg/api/v1/clients/factory/resource_client_factory.go
+++ b/pkg/api/v1/clients/factory/resource_client_factory.go
@@ -6,6 +6,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/crd/client/clientset/versioned"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+
 	"github.com/solo-io/solo-kit/pkg/utils/stringutils"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -60,19 +63,44 @@ func newResourceClient(factory ResourceClientFactory, params NewResourceClientPa
 			return nil, errors.Errorf("must provide a shared cache for the kube resource client")
 		}
 
-		// Validate namespace list:
+		// Validate namespace whitelist:
 		// 1. If no namespace list was provided, default to all namespaces
 		// 2. Error if namespace list contains the empty string plus other values
-		namespaces := opts.NamespaceWhitelist
-		if len(namespaces) == 0 {
-			namespaces = []string{metaV1.NamespaceAll}
+		namespaceWhitelist := opts.NamespaceWhitelist
+		if len(namespaceWhitelist) == 0 {
+			namespaceWhitelist = []string{metaV1.NamespaceAll}
 		}
-		if len(namespaces) > 1 && stringutils.ContainsString(metaV1.NamespaceAll, namespaces) {
+		if len(namespaceWhitelist) > 1 && stringutils.ContainsString(metaV1.NamespaceAll, namespaceWhitelist) {
 			return nil, fmt.Errorf("the kube resource client namespace list must contain either "+
-				"the empty string (all namespaces) or multiple non-empty strings. Found both: %v", namespaces)
+				"the empty string (all namespaces) or multiple non-empty strings. Found both: %v", namespaceWhitelist)
 		}
 
-		return kube.NewResourceClient(opts.Crd, kubeCfg, opts.SharedCache, inputResource, opts.SkipCrdCreation, namespaces, opts.ResyncPeriod)
+		// If the flag is false, call the k8s apiext API to create the given CRD.
+		if !opts.SkipCrdCreation {
+			apiExts, err := clientset.NewForConfig(kubeCfg)
+			if err != nil {
+				return nil, errors.Wrapf(err, "creating api extensions client")
+			}
+			if err := opts.Crd.Register(apiExts); err != nil {
+				return nil, err
+			}
+		}
+
+		// Create clientset for solo resources
+		crdClient, err := versioned.NewForConfig(kubeCfg, opts.Crd)
+		if err != nil {
+			return nil, errors.Wrapf(err, "creating crd client")
+		}
+
+		return kube.NewResourceClient(
+			opts.Crd,
+			crdClient,
+			opts.SharedCache,
+			inputResource,
+			namespaceWhitelist,
+			opts.ResyncPeriod,
+		), nil
+
 	case *ConsulResourceClientFactory:
 		return consul.NewResourceClient(opts.Consul, opts.RootKey, resourceType), nil
 	case *FileResourceClientFactory:
--- a/pkg/api/v1/clients/factory/resource_client_factory.go
+++ b/pkg/api/v1/clients/factory/resource_client_factory.go
@@ -1,8 +1,13 @@
 package factory
 
 import (
+	"fmt"
 	"reflect"
 	"strings"
+	"time"
+
+	"github.com/solo-io/solo-kit/pkg/utils/stringutils"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients/kubesecretplain"
 
@@ -37,8 +42,12 @@ func newResourceClient(factory ResourceClientFactory, params NewResourceClientPa
 	resourceType := params.ResourceType
 	switch opts := factory.(type) {
 	case *KubeResourceClientFactory:
+		kubeCfg := opts.Cfg
+		if kubeCfg == nil {
+			return nil, errors.Errorf("must provide a rest.Config for the kube resource client")
+		}
 		if params.Token != "" {
-			opts.Cfg.BearerToken = strings.TrimPrefix(params.Token, "Bearer ")
+			kubeCfg.BearerToken = strings.TrimPrefix(params.Token, "Bearer ")
 		}
 		inputResource, ok := params.ResourceType.(resources.InputResource)
 		if !ok {
@@ -50,10 +59,19 @@ func newResourceClient(factory ResourceClientFactory, params NewResourceClientPa
 		if opts.SharedCache == nil {
 			return nil, errors.Errorf("must provide a shared cache for the kube resource client")
 		}
-		if opts.Cfg == nil {
-			return nil, errors.Errorf("must provide a resclient.Config for the kube resource client")
+
+		// Validate namespace list:
+		// 1. If no namespace list was provided, default to all namespaces
+		// 2. Error if namespace list contains the empty string plus other values
+		namespaces := opts.Namespaces
+		if namespaces == nil || len(namespaces) == 0 {
+			namespaces = []string{metaV1.NamespaceAll}
+		} else if stringutils.ContainsString(metaV1.NamespaceAll, namespaces) && len(namespaces) > 1 {
+			return nil, fmt.Errorf("the kube resource client namespace list must contain either "+
+				"the empty string (all namespaces) or multiple non-empty strings. Found both: %v", namespaces)
 		}
-		return kube.NewResourceClient(opts.Crd, opts.Cfg, opts.SharedCache, inputResource, opts.SkipCrdCreation)
+
+		return kube.NewResourceClient(opts.Crd, kubeCfg, opts.SharedCache, inputResource, opts.SkipCrdCreation, namespaces, opts.ResyncPeriod)
 	case *ConsulResourceClientFactory:
 		return consul.NewResourceClient(opts.Consul, opts.RootKey, resourceType), nil
 	case *FileResourceClientFactory:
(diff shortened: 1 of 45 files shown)


pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -378,29 +411,42 @@ func (rc *ResourceClient) Watch(namespace string, opts clients.WatchOpts) (<-cha
...
 		defer rc.sharedCache.removeWatch(cacheUpdated)
 		defer close(resourcesChan)
 		defer close(errs)
 
+		// Perform an initial list operation
 		updateResourceList()
 
 		for {
 			select {
 			case <-time.After(opts.RefreshRate): // TODO(yuval-k): can we remove this? is the factory takes care of that...
```
marcogschmidt: @yuval-k I would remove it. This does not really refresh the cache, but merely re-lists its content. The actual resync period (that forces the reflector to check whether the caches are out of sync) can now be specified on the `KubeResourceClientFactory`. Not sure if this is needed somewhere though. In case, the same behavior can be achieved by periodically calling the `List` method on the client itself. What do you think?

pkg/api/v1/clients/kube/controller/controller.go
```diff
@@ -17,132 +17,112 @@ import (
...
+// given informers. The controller also collects the HasSynced functions of each of the informers.
 func NewController(
 	controllerName string,
-	kubeclient kubernetes.Interface,
+	kubeClient kubernetes.Interface,
 	handler cache.ResourceEventHandler,
 	informers ...cache.SharedInformer) *Controller {
 
+	// TODO: how will this work with unauthorized users?
 	eventBroadcaster := record.NewBroadcaster()
```
marcogschmidt: @ilackarms @yuval-k It looks like this `recorder` is currently not used. Right now the controller works and I did not investigate further. I can see how this would fail in case `kubeclient` does not have permissions on `Event`s. What do you think?

pkg/api/v1/clients/kube/resource_client_factory.go
```diff
@@ -64,137 +67,174 @@ type ResourceClientSharedInformerFactory struct {
...
+					defer stats.Record(ctx, MInFlight.M(-1))
+					return watch(options)
+				},
 			},
-		},
-		&v1.Resource{},  // TODO(yuval-k): can we make this rc.crd.Type ?
-		f.defaultResync, // TODO(yuval-k): make this configurable!
-		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
-	)
+			&v1.Resource{}, // TODO(yuval-k): can we make this rc.crd.Type ?
```
marcogschmidt: @yuval-k Not without other changes. Tried just to see if it would work and got this error:
```
ERROR: logging before flag.Parse: E1231 15:42:13.213211    6035 reflector.go:315] github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/controller/controller.go:88: expected type *v1.ChangeSet, but watch event object had type *v1.Resource
```
Thrown from [here](https://github.com/kubernetes/client-go/blob/7856fdbcc3be4d133402982f0a5069d251eb8459/tools/cache/reflector.go#L315).

pkg/api/v1/clients/factory/resource_client_factory.go
```diff
@@ -50,10 +59,19 @@ func newResourceClient(factory ResourceClientFactory, params NewResourceClientPa
...
 			return nil, errors.Errorf("must provide a shared cache for the kube resource client")
 		}
-		if opts.Cfg == nil {
-			return nil, errors.Errorf("must provide a resclient.Config for the kube resource client")
+
+		// Validate namespace list:
+		// 1. If no namespace list was provided, default to all namespaces
+		// 2. Error if namespace list contains the empty string plus other values
+		namespaces := opts.Namespaces
+		if namespaces == nil || len(namespaces) == 0 {
```
ilackarms: nit: `len(namespaces) == 0` covers the `namespaces == nil`

https://golang.org/pkg/builtin/#len

pkg/api/v1/clients/factory/resource_client_factory.go
```diff
@@ -50,10 +59,19 @@ func newResourceClient(factory ResourceClientFactory, params NewResourceClientPa
...
+			namespaces = []string{metaV1.NamespaceAll}
+		} else if stringutils.ContainsString(metaV1.NamespaceAll, namespaces) && len(namespaces) > 1 {
```
ilackarms: nit: remove `else`, change to 
`if stringutils.ContainsString(metaV1.NamespaceAll, namespaces) {` 
(if `len(namespaces) == 0` then `stringutils.ContainsString(metaV1.NamespaceAll, namespaces)` will always be false)

pkg/api/v1/clients/factory/resource_client_factory.go
```diff
@@ -81,11 +99,15 @@ type ResourceClientFactory interface {
...
 // during registration. This allows us to create and register resource clients in cases where the given configuration
 // contains a token associated with a user that is not authorized to create CRDs.
+// Clients built with this factory will be able to access only resources the given namespace list. If no value is provided,
+// clients will be able to access resources in all namespaces.
 type KubeResourceClientFactory struct {
 	Crd             crd.Crd
 	Cfg             *rest.Config
 	SharedCache     *kube.KubeCache
 	SkipCrdCreation bool
+	Namespaces      []string
```
ilackarms: suggestion: replace `Namespaces` with something like `NamespaceWhitelist` so it's more clear what function the list serves

pkg/api/v1/clients/kube/resource_client.go
ilackarms: the idea is that this helps "kick the tires" to make sure event loops get started immediately upon calling a watch. if you noticed this function isn't necessary, maybe because a watch event happens immediately anyway, we could remove it. but i think removing it might disturb some of our event loops (in particular i believe function discovery benefitted from this)

pkg/api/v1/clients/kube/resource_client_factory.go
```diff
@@ -64,137 +67,174 @@ type ResourceClientSharedInformerFactory struct {
 	lock          sync.Mutex
 	defaultResync time.Duration
 
-	informers map[reflect.Type]cache.SharedIndexInformer
-	// startedInformers is used for tracking which informers have been started.
+	informers map[reflect.Type]map[string]cache.SharedIndexInformer
```
ilackarms: maybe consider converting this `map[reflect.Type]map[string]cache.SharedIndexInformer` to some kind of struct for easier management? it''s not terrible, but having a struct enclosing these maps with some `Add` `Remove` functions can simplify this a bit

pkg/api/v1/clients/kube/resource_client_factory.go
ilackarms: can you explain a bit more? is this still an issue?

ilackarms:
overall looks good, i gave some style suggestions 

my only concern is that this is a change to a fairly complicated (and not well-tested) piece of the code. how hard do you think it would be to add some test coverage around this code?

pkg/api/v1/clients/kube/resource_client.go
marcogschmidt: I think you might be referring to line 420? I was talking about the `TODO` @yuval-k left at line 424 (the timeout clause in the `select` statement).

pkg/api/v1/clients/kube/controller/controller.go
marcogschmidt: Ok, will do

pkg/api/v1/clients/kube/resource_client_factory.go
marcogschmidt: I was replying to the second `TODO` comment. The parameter is used by the reflector to check the type of the event object it receives from the `ListWatch`. By passing `&v1.Resource{}` it works, but if you pass `rc.crd.Type` you get the error I mentioned. 

I believe the parameter must match the resource type that the `WatchFunc` in line 121
```go
watch := rc.kube.ResourcesV1().Resources(ns).Watch
```
is defined for, if that makes sense. Yuval was probably wondering if by passing `rc.crd.Type` you could get type information with the event for free, but from what I was able to tell there is no way of getting anything more than our `*v1.Resource`.

pkg/api/v1/clients/factory/resource_client_factory.go
marcogschmidt: Will change it

pkg/api/v1/clients/factory/resource_client_factory.go
marcogschmidt: Good point, will do

pkg/api/v1/clients/kube/controller/controller.go
```diff
@@ -17,132 +17,112 @@ import (
...
+// Starts the controller by:
+//
+// 1. Registering the event handler with each if the informers
+// 2. Starting each informer
+// 3. Wait for the informer caches to sync
+// 4. Starting a number of parallel workers equal to the "parallelism" parameter
+//
+// When a message is received on the given stopCh, the controller will stop the informers, shutdown the work queue and
+// wait for workers to finish processing their current work items.
+func (c *Controller) Run(parallelism int, stopCh <-chan struct{}) error {
```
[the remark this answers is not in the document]
marcogschmidt: 😊

pkg/api/v1/clients/kube/resource_client_factory.go
marcogschmidt: Makes a lot of sense, I'll do it

pkg/api/v1/clients/kube/controller/controller.go
```diff
@@ -17,132 +17,112 @@ import (
...
+	// recorder is an event recorder for recording Event resources to the Kubernetes API
 	recorder record.EventRecorder
 
 	// handler to call
 	handler cache.ResourceEventHandler
 }
 
-// NewController returns a new controller
+// Returns a new kubernetes controller without starting it. The given event handler is registered with each of the
+// given informers. The controller also collects the HasSynced functions of each of the informers.
```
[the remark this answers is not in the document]
marcogschmidt: Good catch, will update the comment

pkg/api/v1/clients/kube/controller/controller.go
```diff
@@ -17,132 +17,112 @@ import (
...
 
-	log.Debugf("Started workers")
-	<-stopCh
-	log.Debugf("Shutting down workers")
+	// Start workers in goroutine so we can defer the queue shutdown
+	go func() {
+		defer c.workQueue.ShutDown()
+		log.Debugf("Starting workers")
+
+		// Launch two workers to process resources
```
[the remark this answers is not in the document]
marcogschmidt: Fixed

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -162,13 +159,18 @@ type ResourceClient struct {
...
 
-func NewResourceClient(crd crd.Crd, cfg *rest.Config, sharedCache *KubeCache, resourceType resources.InputResource, skipCrdCreation bool) (*ResourceClient, error) {
+func NewResourceClient(crd crd.Crd, cfg *rest.Config, sharedCache *KubeCache, resourceType resources.InputResource,
+	skipCrdCreation bool, namespaces []string, resyncPeriod time.Duration) (*ResourceClient, error) {
+
 	apiExts, err := apiexts.NewForConfig(cfg)
 	if err != nil {
 		return nil, errors.Wrapf(err, "creating api extensions client")
 	}
+	//
```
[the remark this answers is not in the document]
marcogschmidt: Done

marcogschmidt:
Summary of the commits since the last review: 
- added tests fo
  - k8s controller
  - shared cache
  - kube resource client
- simplified shared cache design and resolved bugs that were surfaced by the tests
- configured CI to run all test suites
  - NOTE: I had to comment out a lot of tests since they were failing and fixing them is beyond the scope of this PR

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -311,16 +248,23 @@ func (rc *ResourceClient) Delete(namespace, name string, opts clients.DeleteOpts
...
-	rc.sharedCache.startFactory(context.TODO(), rc.kubeClient)
 
-	lister, err := rc.sharedCache.sharedInformerFactory.GetLister(rc.crd.Type)
+	namespace = clients.DefaultNamespaceIfEmpty(namespace)
+	if err := rc.validateNamespace(namespace); err != nil {
+		return nil, err
+	}
+
+	// Will have no effect if the factory is already running
+	rc.sharedCache.Start(context.TODO())
```
ilackarms: do we want to use `opts.WatchOpts.Ctx` here? @yuval-k

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -356,10 +300,15 @@ func (rc *ResourceClient) List(namespace string, opts clients.ListOpts) (resourc
...
 func (rc *ResourceClient) Watch(namespace string, opts clients.WatchOpts) (<-chan resources.ResourceList, <-chan error, error) {
-	rc.sharedCache.startFactory(context.TODO(), rc.kubeClient)
 
-	opts = opts.WithDefaults()
 	namespace = clients.DefaultNamespaceIfEmpty(namespace)
+	if err := rc.validateNamespace(namespace); err != nil {
+		return nil, nil, err
+	}
+
+	rc.sharedCache.Start(context.TODO())
```
ilackarms: do we want to use opts.WatchOpts.Ctx here? @yuval-k

pkg/api/v1/clients/kube/resource_client.go
yuval-k: In theory, as long as the credentials do not change the cache neednt expire.
Assuming opts.WatchOpts.Ctx expires on settings change which should be rare, i don't see a big different either way for performance. the question is of correctness:

is Start be called again when that context is canceled? what is the lifetime of sharedCache?

marcogschmidt: That's a good point. Right now the controller (informers + workers) is stopped when the context that you pass to `Start(ctx)` is cancelled., which is never, since we are passing a todo context.

The most correct behavior would probably be to get the context in the resourceClient constructor and to propagate it to the controller. That way client code can determine their own lifecycle for the resourceClient.

yuval-k: it's fine that they have never stopped, as long as it is consistent - 
i.e. when the next start is called it will re-use the existing object?

marcogschmidt: After you call `start` the first time, subsequent invocations do not have any effect, and you can't register clients after the cache has been started. So there is no way to reuse it. You have to register all the clients that want to use the cache beforehand and then start it. After that it just lives forever. 

You're right that it's probably not a problem, given that AFAIK in our applications the client lifetime is the same as the one of the application, so there's no risk of leaking resources. Less than ideal, but good for now I guess.

pkg/api/v1/clients/kube/resource_client_factory.go
```diff
@@ -54,145 +56,248 @@ func init() {
...
+	defer f.cacheUpdatedWatchersMutex.Unlock()
+	for _, watcher := range f.cacheUpdatedWatchers {
+		// Attempt delivery asynchronously
+		go func(watchChan chan v1.Resource, res v1.Resource) {
+			select {
+			case watchChan <- resource:
+			case <-time.After(f.watchTimeout):
+				contextutils.LoggerFrom(context.TODO()).Errorf("timed out after waiting for %v"+
+					"for watch to receive event on resource %v", f.watchTimeout, resource)
+			}
```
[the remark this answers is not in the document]
marcogschmidt: Any comment about this? @ilackarms @yuval-k

pkg/api/v1/clients/kube/resource_client.go
marcogschmidt: Maybe we could add a second constructor (not to break backwards-compatibility) in the form:
```go
NewKubeCache(done <-chan struct{})
```
That would be pretty quick.

yuval-k: i would worry about backward-compatibility as there is no stable release.
lets talk about tomorrow morning - as I'm trying to understand why it doesn't work now vs before the change.

marcogschmidt: Nothing really changed since the prior version, we already were passing a `ctx.TODO()` to the factory and never shutting it down (see [current `master`](https://github.com/solo-io/solo-kit/blob/bd125b289e404433f59b32148cb6eb54c488c5ec/pkg/api/v1/clients/kube/resource_client.go#L321)).

PR #55, opened 2019-01-04
Resource groups from config

ilackarms:
allow resource groups to be specified in `solo-kit.json` rather than relying on magic comments in the protos

allows event loops to be generated for your application without needing to modify an external proto file


ilackarms:
just a note: this was tested against (and is a dependency of) the ingress PR on gloo: https://github.com/solo-io/gloo/pull/212

code generation is working well; the "legacy" method is still supported

PR #57, opened 2019-01-04
simplify defer call

mitchdraft:
Removes redundant function body

This change is valid because from https://golang.org/ref/spec#Defer_statements cancel's arguments (there are no arguments in this case) will be evaluated when defer executes but cancel will not be evaluated until the parent function exits

I tested by:
- making the equivalent change to squash's solo-kit artifacts and running its tests
- making the equivalent changes to gloo's api and setup event loops and running gloo

### DIFF
--- a/pkg/code-generator/codegen/templates/event_loop_template.go
+++ b/pkg/code-generator/codegen/templates/event_loop_template.go
@@ -67,7 +67,8 @@ func (el *{{ lower_camel .GoName }}EventLoop) Run(namespaces []string, opts clie
 	go func() {
 		// create a new context for each loop, cancel it before each loop
 		var cancel context.CancelFunc = func() {}
-		defer cancel()
+		// use closure to allow cancel function to be updated as context changes
+		defer func() { cancel() }()
 		for {
 			select {
 			case snapshot, ok := <-watch:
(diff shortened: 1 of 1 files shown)


mitchdraft:
@yuval-k could you comment on the need for the prior form?
I suspect that I need to revert my first commit (I didn't realize that cancel was overwritten) - if so I'll use this pr to add a comment

pkg/code-generator/codegen/templates/event_loop_template.go
```diff
@@ -67,7 +67,7 @@ func (el *{{ lower_camel .GoName }}EventLoop) Run(namespaces []string, opts clie
 	go func() {
 		// create a new context for each loop, cancel it before each loop
 		var cancel context.CancelFunc = func() {}
-		defer func() { cancel() }()
+		defer cancel()
```
yuval-k: the reason i used the more complicated syntax is because cancel is a variable,
and i want to make sure that defer uses it's latest value, and not its current value (as its value is changed in the select block.

I'm not 100% sure of the defer semantics, and there's a chance that this will work, I'm ok with this change if you can add a unit tests that proves that this works

mitchdraft:
The prior structure is needed as Yuval demonstrated with https://play.golang.org/p/2tWg3AEeGGs

PR #61, opened 2019-01-08
Test template cache fix

ilackarms:
this fixes a generated test: 
 this change checks ahead-of-time whether any resource does *not* have `HasStatus` set. if not, it creates a kube clientset. if any resource `HasStatus`, it creates a kube cache. if the resource group has both, the test needs both

### DIFF
--- a/pkg/code-generator/codegen/templates/snapshot_emitter_test_template.go
+++ b/pkg/code-generator/codegen/templates/snapshot_emitter_test_template.go
@@ -70,7 +70,9 @@ var _ = Describe("{{ upper_camel .Project.ProjectConfig.Version }}Emitter", func
 		cfg, err = clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 		Expect(err).NotTo(HaveOccurred())
 
+{{- if $needs_kube_client }}
 		cache := kuberc.NewKubeCache()
+{{- end}}
 
 {{- if $needs_kube_client }}
 var kube kubernetes.Interface
--- a/pkg/code-generator/codegen/templates/snapshot_emitter_test_template.go
+++ b/pkg/code-generator/codegen/templates/snapshot_emitter_test_template.go
@@ -8,13 +8,10 @@ var ResourceGroupEmitterTestTemplate = template.Must(template.New("resource_grou
 
 {{- /* we need to know if the tests require a crd client or a regular clientset */ -}}
 {{- $clients := new_str_slice }}
-{{- $needs_crd_client := false }}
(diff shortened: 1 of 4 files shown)


pkg/code-generator/codegen/templates/snapshot_emitter_test_template.go
```diff
@@ -67,11 +70,14 @@ var _ = Describe("{{ upper_camel .Project.ProjectConfig.Version }}Emitter", func
...
 
+{{- if $needs_crd_client }}
+
 		cache := kuberc.NewKubeCache()
+{{- end}}
 
-{{- if $needs_kube_client }}
-var kube kubernetes.Interface
-{{- end }}
+{{- if $needs_standard_client }}
```
rickducott: Seems a bit weird to use `needs_standard_client` and `need_crd_client` here, but not refer to them below.

pkg/code-generator/codegen/templates/snapshot_emitter_test_template.go
```diff
@@ -70,11 +73,14 @@ var _ = Describe("{{ upper_camel .Project.ProjectConfig.Version }}Emitter", func
 		cfg, err = clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 		Expect(err).NotTo(HaveOccurred())
 
+{{- if $needs_crd_client }}
+
 		cache := kuberc.NewKubeCache()
+{{- end}}
```
marcogschmidt: I think you can just call `NewKubeCache()` on line 92. Since it's needed only there, and given that you land in that block only if `.HasStatus` is true, `needs_crd_client` becomes redundant and can be removed.

pkg/code-generator/codegen/templates/snapshot_emitter_test_template.go
```diff
@@ -70,11 +73,14 @@ var _ = Describe("{{ upper_camel .Project.ProjectConfig.Version }}Emitter", func
...
 
-{{- if $needs_kube_client }}
-var kube kubernetes.Interface
-{{- end }}
+{{- if $needs_standard_client }}
+		var kube kubernetes.Interface
+{{- end}}
```
marcogschmidt: Instead of declaring it here, you could use `:=` instead of `=` on line 95. It's used only there.

ilackarms: i can't because if there are multiple resources that use it, we'll have a redeclare error:
```
kube, err := ...
....
kube, err := ... < -- error here
```

PR #81, opened 2019-02-04
NamespaceAll, Secret/Configmap Caching

ilackarms:
this PR does 3 changes:

- support using `NamespaceAll` in the 3 different kube clients (crd, secret, configmap)
- adds use of a kubecontroller-backed cache to the secret and configmap clients (currently they're just using List in a refresh loop)
- merge the kubesecret client and kubesecretplain client
    - kubesecretplain was created during supergloo development, the only difference it has with the regular client is how resources are marshalled/unmarshalled
    - this was done to reduce the amount of copypasting going on
    - tests have been preserved


pkg/api/v1/clients/configmap/resource_client_test.go
```diff
@@ -71,4 +79,109 @@ var _ = Describe("Base", func() {
...
+			ns1, ns2 string
+			cfg      *rest.Config
+			client   *ResourceClient
+		)
+		BeforeEach(func() {
+			ns1 = helpers.RandString(8)
+			ns2 = helpers.RandString(8)
+			err := setup.SetupKubeForTest(ns1)
+			Expect(err).NotTo(HaveOccurred())
+			err = setup.SetupKubeForTest(ns2)
```
marcogschmidt: Check error

pkg/api/v1/clients/kubesecret/resource_client_test.go
```diff
@@ -38,12 +45,118 @@ var _ = Describe("Base", func() {
...
+			ns1, ns2 string
+			cfg      *rest.Config
+			client   *ResourceClient
+		)
+		BeforeEach(func() {
+			ns1 = helpers.RandString(8)
+			ns2 = helpers.RandString(8)
+			err := setup.SetupKubeForTest(ns1)
+			Expect(err).NotTo(HaveOccurred())
+			err = setup.SetupKubeForTest(ns2)
```
marcogschmidt: Check error

pkg/api/v1/clients/kubesecret/resource_client.go
```diff
@@ -81,19 +80,70 @@ func (rc *ResourceClient) toKubeSecret(resource resources.Resource) (*v1.Secret,
 	}, nil
 }
 
+func (rc *ResourceClient) fromPlainKubeSecret(secret *v1.Secret) (resources.Resource, error) {
+	resource := rc.NewResource()
+	// not our secret
+	// should be an error on a Read, ignored on a list
+	if len(secret.ObjectMeta.Annotations) == 0 || secret.ObjectMeta.Annotations[annotationKey] != rc.Kind() {
```
marcogschmidt: Would it make sense to maybe log something here? Just so that we are aware of when these unexpected cases occur.

pkg/api/v1/clients/kubesecret/resource_client.go
```diff
@@ -81,19 +80,70 @@ func (rc *ResourceClient) toKubeSecret(resource resources.Resource) (*v1.Secret,
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "marshalling resource as map")
+	}
+	kubeSecretData := make(map[string][]byte)
+	for k, v := range resourceMap {
+		switch val := v.(type) {
+		case string:
+			kubeSecretData[k] = []byte(val)
+		default:
+			// TODO: handle other field types; for now the caller
```
marcogschmidt: Can log a warning here to be aware of when this happens?

pkg/api/v1/clients/kubesecret/resource_client.go
marcogschmidt: Ok, makes sense

Issue #70, opened 2019-01-22
watch all namespaces


mitchdraft:
Add a mode to watch all namespaces

Should also be able to add/remove namespaces from the watch when they are created/deleted

This will involve creating a way to create a new event loop and drain the old event loop

yuval-k:
this is fixed by #82 and #81 
@ilackarms correct me if i'm wrong

PR #82, opened 2019-02-06
support "" (AllNamespaces) in snapshot emitter / event loop

ilackarms:
this PR adds the ability to pass `nil` or `[]string{""}` to the EventLoop in Solo-Kit

before: if a watch was started on namespace `""` and produced a list of resources from multiple namespaces, the snapshot produced would have a key in `Snapshot.{{Resource.Name}}ByNamespace` for each namespace. The problem is that when the same watch produced a new list of resources, we don't know which resources to clear from the old snapshot.

now: when a new resource list comes from a watch, the key in `Snapshot.{{Resource.Name}}ByNamespace` will be the original namespace that started the watch (i.e. `""`).

event loops still support passing in a whitelist of namespaces, but `nil` is now a valid argument which indicates "watch all namespaces"


yuval-k:
this fixes https://github.com/solo-io/solo-kit/issues/70
and related to (fixes?) https://github.com/solo-io/gloo/issues/292

PR #87, opened 2019-02-20
for eds to work, mem client need to support all namespaces

yuval-k:


ilackarms:
can we add a unit test for this? otherwise lgtm

yuval-k:
i modified the existing test

ilackarms:
@yuval-k can you explain the modification?

yuval-k:
yes, list and watch will be done with an empty namespace now and thus will only work if all namespaces work

ilackarms:
i would prefer not to break these tests for file, vault, grpc and consul clients

instead of modifying these tests, can you add a test specifically for the empty namespace case for this client type? there are example tests you can copy-paste for this already:
* https://github.com/solo-io/solo-kit/blob/master/pkg/api/v1/clients/kube/resource_client_test.go#L148
* https://github.com/solo-io/solo-kit/blob/master/pkg/api/v1/clients/kubesecret/resource_client_test.go#L88
* https://github.com/solo-io/solo-kit/blob/master/pkg/api/v1/clients/configmap/resource_client_test.go#L111

yuval-k:
well the tests pass so nothing seems to be broken...

ilackarms:
consul and vault tests are disabled in CI

additionally i find the changes to the `test_crud_client` helper function to be a bit confusing (i.e what's the difference between `namespace` and `writeNamespace`?). are you opposed to adding a separate test for this case?

yuval-k:
i think that consul and vault should be changed to support all namespaces. so i think changing this test is the correct thing. and we can fix consul client in a follow-up PR; thoughts?

PR #103, opened 2019-03-13
camel case field names in docs, regen docs

### LINKED ISSUE #102: autogenerated API docs should output camel, not snake case
Our protos are written in snake case, but the yaml is persisted in Kube as camel case. Because the docs are written based on the protos, they are persisted as snake case. This makes it confusing to read the docs and try to update kube config via yaml.

ilackarms:
fixes #102


rickducott:
Updated generated docs look good

PR #104, opened 2019-03-14
dont start informers twice in cache.go

yuval-k:


pkg/api/v1/clients/kube/controller/controller.go
```diff
@@ -32,14 +33,15 @@ type Controller struct {
 // Returns a new kubernetes controller without starting it.
 func NewController(
 	controllerName string,
-	handler cache.ResourceEventHandler,
+	handler cache.ResourceEventHandler, RunInformers bool,
```
ilackarms: i'd suggest an alternative to passing a bool from various places in the code. there's no indication to clients what `RunInformers` will do, nor that the system will break if more than one client passes `true` here. I think the behavior of ensuring caches are only started once should be deferred to something like a `sync.Once`, or by ensuring that only a single reference to the `Controller` is shared by different clients

yuval-k: it not possible as the other object that starts them is the shared informer factory that doesnt use the our controller (as it is kubernetes code).
this patch compiles with gloo - i.e. in gloo the controller is only used via solo-kit and no other change is needed.

ilackarms: in the case that a user is using only the `KubeCoreCache` without the CRD `KubeCache`, this approach will fail. for this reason among others i think it is fragile to rely on callers being synchronized without providing any way coordinating the starting of Informers across callers.

wouldn't this be a problem if callers are using different sets of informers and only one set is started? 

why is it not possible ensure informers are only started once via a `sync.Once`? it seems to me that we need a global `Once` for each type of informer we plan to start. the issue gets even more complicated in the case of the apiserver where each token gets its own cache

pkg/api/v1/clients/kube/cache/cache.go
```diff
@@ -52,7 +52,7 @@ func NewKubeCoreCache(ctx context.Context, client kubernetes.Interface) (*KubeCo
 	}
 
 	kubeController := controller.NewController("kube-plugin-controller",
-		controller.NewLockingSyncHandler(k.updatedOccured),
+		controller.NewLockingSyncHandler(k.updatedOccured), false,
 		pods.Informer(), configMaps.Informer(), secrets.Informer())
 
 	stop := ctx.Done()
```
marcogschmidt: Instead of adding that flag, can't we just remove the `go kubeInformerFactory.Start(stop)` line? If you look at that function, it just starts all the informers in the factory: 
```go
// Start initializes all requested informers.
func (f *sharedInformerFactory) Start(stopCh <-chan struct{}) {
	f.lock.Lock()
	defer f.lock.Unlock()

	for informerType, informer := range f.informers {
		if !f.startedInformers[informerType] {
			go informer.Run(stopCh)
			f.startedInformers[informerType] = true
		}
	}
}
```
That is the same thing that our kube controller would do if you passed it the same informers, which we do. 

Basically, it seems to me that all we need the `kubeInformerFactory` for is initializing the core informers via those `kubeInformerFactory.Core().V1().*` functions.

yuval-k: that is a good idea; i didnt noticed that the sharedInformerFactory isnt really used last night.
i saw the code now and indeed the only thing we do is call `Start()` so we can just not do that.

yuval-k: you can skip calling `kubeInformerFactory.Start` but still call the `kubeController.WaitForInformers` I added in this PR to make it more foolproof (instead of getting the list of informers again)

pkg/api/v1/clients/kube/controller/controller.go
yuval-k: In general global state is not a good idea - It is indeed impossible to have a per token cache with a global `Once`.
I think marco suggestion below should work.

ilackarms: i like it

yuval-k:
looks good to me

PR #108, opened 2019-03-20
kubesecret: allow using the default behavior from a secret converter

yuval-k:


pkg/api/v1/clients/kubesecret/resource_client.go
```diff
@@ -83,7 +83,9 @@ func (rc *ResourceClient) ToKubeSecret(ctx context.Context, resource resources.R
 }
 
 type SecretConverter interface {
+	// If this methoed returns nil,nil the default conversion will be used.
```
ilackarms: method

Issue #109, opened 2019-03-20
plain configmaps client


ilackarms:
for supergloo, need a way to read/write a prometheus configmap yaml as a  `map[string]string`

this works for plainsecrets, but not (yet) for configmaps

PR #110, opened 2019-03-20
plain configmap client

ilackarms:
allow storing configmaps as map[string]string #109

### DIFF
--- a/pkg/api/v1/clients/configmap/converters.go
+++ b/pkg/api/v1/clients/configmap/converters.go
@@ -1,0 +1,118 @@
+package configmap
+
+import (
+	"context"
+
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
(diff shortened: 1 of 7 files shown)


ilackarms:
/kick
dep failed with 502

pkg/api/v1/clients/configmap/converters.go
```diff
@@ -0,0 +1,118 @@
...
+		if k == "metadata" {
+			continue
+		}
+		switch val := v.(type) {
+		case string:
+			configMapData[k] = val
+		default:
+			// TODO: handle other field types; for now the caller
+			// must know this resource client only supports map[string]string style objects
+			contextutils.LoggerFrom(ctx).Warnf("invalid resource type (%v) used for plain configmap. unable to " +
```
marcogschmidt: We are missing the operand for the `%v` format specifier. Can you fix this in the secret client as well?

PR #113, opened 2019-03-20
fix return nil semantics add specific error for a resource that doesnt belong to us

yuval-k:


pkg/api/v1/clients/kubesecret/resource_client.go
```diff
@@ -23,12 +24,16 @@ import (
 
 const annotationKey = "resource_kind"
 
+var (
+	NotOurResource error = fmt.Errorf("not kube secret resource")
```
ilackarms: make private, else lgtm

yuval-k: i want this public to potentially be used from the SecretConverter interface

PR #115, opened 2019-03-21
Fix docs links

rickducott:


rickducott:
This is working now, tested with https://github.com/solo-io/solo-docs/pull/196, which simulates the PR created by gloo releases (note: i only tested with gloo docs, not solo-kit or solo-projects, so not EVERY link is fixed on that branch). 

Basically, when rendering Hugo anchor links, we can use the simplest Hugo convention: lower-case heading name. Attempted to not change non-Hugo style docs. Since our anchor links all point to type names, they don't include spaces and I didn't implement converting spaces to dashes.

rickducott:
Once this merges, I will release, then we can start upgrading projects and releasing them to fix the links.

christian-posta:
Related to https://github.com/solo-io/solo-docs/issues/170

PR #116, opened 2019-03-21
use go utils instead of solo-kit for context utils

yuval-k:


rickducott:
Can you add a breaking changelog? I like this change, looks good.

PR #121, opened 2019-04-01
Support special INVALID state for objects in snapshot

rickducott:


pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -377,6 +392,20 @@ func (rc *ResourceClient) exist(ctx context.Context, namespace, name string) boo
 
 }
 
+func (rc *ResourceClient) convertCrdToInvalidResource(resourceCrd *v1.Resource) (resources.Resource, error) {
+	resourceCrd.Status.State = core.Status_Invalid
+	_, err := rc.crdClientset.ResourcesV1().Resources(resourceCrd.Namespace).Update(resourceCrd)
+	if err != nil {
+		return nil, err
```
yuval-k: if any of these errors happen we need to make sure that we don't update the snapshot.

i.e. we must guarantee that if an object exists in storage, it will also exist in the snapshot.

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -377,6 +392,20 @@ func (rc *ResourceClient) exist(ctx context.Context, namespace, name string) boo
 
 }
 
+func (rc *ResourceClient) convertCrdToInvalidResource(resourceCrd *v1.Resource) (resources.Resource, error) {
+	resourceCrd.Status.State = core.Status_Invalid
+	_, err := rc.crdClientset.ResourcesV1().Resources(resourceCrd.Namespace).Update(resourceCrd)
```
yuval-k: only do an update if we have to. i.e. the status != invalid

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -282,7 +294,10 @@ func (rc *ResourceClient) List(namespace string, opts clients.ListOpts) (resourc
 	for _, resourceCrd := range listedResources {
 		resource, err := rc.convertCrdToResource(resourceCrd)
 		if err != nil {
-			return nil, errors.Wrapf(err, "converting output crd")
+			resource, err = rc.convertCrdToInvalidResource(resourceCrd)
```
ilackarms: i think we should propagate the conversion error here. set the `reason` field on the Status so the user knows where parsing failed

pkg/api/v1/clients/kube/resource_client.go
ilackarms: reporting the status to storage should probably happen in another function

also, that function should be gated behind a parameter for the constructor like `EnableErrorReporting`

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -377,6 +392,20 @@ func (rc *ResourceClient) exist(ctx context.Context, namespace, name string) boo
...
 
+func (rc *ResourceClient) convertCrdToInvalidResource(resourceCrd *v1.Resource) (resources.Resource, error) {
+	resourceCrd.Status.State = core.Status_Invalid
+	_, err := rc.crdClientset.ResourcesV1().Resources(resourceCrd.Namespace).Update(resourceCrd)
+	if err != nil {
+		return nil, err
+	}
+	updatedResource := rc.NewInvalidResource()
+	if updatedResource == nil {
+		return nil, errors.Errorf("Could not handle invalid resource %v", resourceCrd)
```
ilackarms: remove nil` err check here, should never happen

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -134,6 +134,18 @@ func (rc *ResourceClient) NewResource() resources.Resource {
 	return resources.Clone(rc.resourceType)
 }
 
+func (rc *ResourceClient) NewInvalidResource() resources.Resource {
+	res, ok := resources.Clone(rc.resourceType).(resources.InputResource)
+	if !ok {
+		return nil
+	}
```
ilackarms: suggestion: this part of the code can call `rc.NewResource` and skip the type check
also allows skipping the `nil` check later on

rickducott: Don't I need the type check to access the status functions?

rickducott: Nvm, I see now

Issue #123, opened 2019-04-04
implement support for custom types in solo-kit


ilackarms:
this would make solo-kit able to support a generic go struct that implements the `Resource` interface to utilize resource clients and event loops.

PR #124, opened 2019-04-04
Custom types support

ilackarms:
Adds the ability to use custom structs with solo kit. struct must implement `resources.Resource` interface
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/123


pkg/api/v1/clients/memory/resource_client.go
```diff
@@ -162,17 +161,14 @@ func (rc *ResourceClient) Write(resource resources.Resource, opts clients.WriteO
...
+	resource.SetMetadata(meta)
 
-	rc.cache.Set(key, clone)
+	rc.cache.Set(key, resource)
 
-	// return another clone!
-	clone = proto.Clone(clone).(resources.Resource)
-
-	return clone, nil
+	return resource, nil
```
yuval-k: why not return another clone? this is important so that modfications on the returned object will not changed what's stored inside.

pkg/api/v1/resources/resource_interface.go
```diff
@@ -15,12 +15,27 @@ import (
...
+	proto.Message
+}
+
+func ProtoCast(res Resource) (ProtoResource, error) {
+	if res == nil {
+		return nil, nil
+	}
+	protoResource, ok := res.(ProtoResource)
+	if !ok {
+		return nil, errors.Errorf("internal server error: unexpected type %T not convertible to resources.Proto", res)
```
yuval-k: internal server error?

pkg/api/v1/resources/resource_interface.go
```diff
@@ -309,8 +324,19 @@ type HashableResource interface {
...
+
 func Clone(resource Resource) Resource {
-	return proto.Clone(resource).(Resource)
+	if cloneable, ok := resource.(CloneableResource); ok {
+		return cloneable.Clone()
+	}
+	if protoMessage, ok := resource.(ProtoResource); ok {
+		return proto.Clone(protoMessage).(Resource)
+	}
+	return nil
```
yuval-k: return nil here is a bit concerning;
let's discuss

pkg/code-generator/codegen/generator.go
```diff
@@ -25,7 +25,7 @@ func GenerateFiles(project *model.Project, skipOutOfPackageFiles bool) (code_gen
 	}
 	for _, res := range project.Resources {
 		// only generate files for the resources in our group, otherwise we import
-		if res.ProtoPackage != project.ProtoPackage {
+		if res.ProtoPackage != project.ProtoPackage && !res.IsCustom {
```
yuval-k: not sure i understand why this is needed - let's discuss

pkg/api/v1/clients/memory/resource_client.go
ilackarms: it already is cloned on line 164

PR #125, opened 2019-04-06
export list error string

mitchdraft:


mitchdraft:
this produced a lot of generated docs unrelated to the change
for clarity, I split out the regen from the new feature in a new PR: https://github.com/solo-io/solo-kit/pull/126

Issue #122, opened 2019-04-02
export common list errors


mitchdraft:
Very minor suggestion:

export errors strings like this:

```go
const UpstreamListErrorTag = "list did not find upstream"
```

as was done in:
https://github.com/solo-io/gloo/pull/651/files#diff-8eec5c179347890490d95491acb44602R3

what do you think @ilackarms ?
seems like it *could* be useful

I will implement if you think it's worth having

ilackarms:
@mitchdraft i don't see any problem with this. what use case did you have in mind?

mitchdraft:
@ilackarms the use case is better error classification for human-friendly output on a cli tool, like this: https://github.com/solo-io/gloo/blob/master/projects/gloo/cli/pkg/printers/virtualservice.go#L84

I exported the string of interest by hand for now: https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/api/v1/upstream_mod.go
but it'd be better for solo-kit to provide it.

one other potential use case in `gloo`:
```
projects/gloo/pkg/translator/translator_test.go
380:                    Expect(err.Error()).To(ContainSubstring("destination # 1: upstream not found: list did not find upstream gloo-system.notexist"))
```

ilackarms:
there's no guarantee that an error containing this string is indeed the type of error we match it to. What about generating an error type instead of a const string? we can have that error type returned, and even provide a util function to assert on the error type:

```golang
type myTypeNotFoundErr struct {
   msg string
}

func IsMyTypeNotFound(err error) bool {
   _, ok := err.(*myTypeNotFoundErr)
   return ok
}

```

this also prevents us from needing to search strings, which can get messy

mitchdraft:
That's a good idea and I would prefer it if I were reading the error from the function that produced it but in this case I'm checking the string that is stored in the `vs.Status.SubresourceStatuses ` field
I agree that it's messy

ilackarms:
sounds like a fairly specific case. do you think it warrants generating additional code, or is it something that we can handle further up the stack (closer to where it is needed)? 

i'm just trying to evaluate the usefulness / potential confusion of having a `v1.<Proxy,Upstream,Secret,Artifact...>ErrorTag` being generated without any clear reason to someone reading the package of what it might be used for

mitchdraft:
yeah, doesn't seem worth it

PR #126, opened 2019-04-06
Export list error reordered commits

mitchdraft:


mitchdraft:
@ilackarms a lot of the diffs have nothing to do with this change, so I reworked the commit order from https://github.com/solo-io/solo-kit/pull/125 and now:
- 1st commit regens resources from unchanged master repo after dep ensure
- 2nd commit `apply change: export list error as string` has new content - please direct your attention there

ilackarms:
@mitchdraft this looks fine to me, but wondering what purpose this change is for?

mitchdraft:
slightly better error handling for user-facing cli output https://github.com/solo-io/solo-kit/issues/122#issuecomment-480489370

mitchdraft:
if we find more of a need for this in the future, we can add it then. Utility seems too limited to add to all projects

PR #128, opened 2019-04-08
check err on unmarshal

mitchdraft:


rickducott:
Note that if you try to release master after this merges in, you will fail to release, because there won't be any changelog for v0.4.3. Consider adding a changelog entry, even if NON_USER_FACING

PR #152, opened 2019-04-26
add cluster to metadata, implement cluster client wrapper

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/151

PR #153, opened 2019-04-26
turns a stream of v1.KubeConfigs into callbacks for Adding/Removing clusters

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/152


pkg/multicluster/config_handler.go
```diff
@@ -0,0 +1,118 @@
...
+	if local != nil {
+		cfgs[LocalCluster] = local
+	}
+
+	for _, kc := range kcs {
+		raw, err := clientcmd.Write(kc.Config)
+		if err != nil {
+			return nil, err
+		}
+		restCfg, err := clientcmd.RESTConfigFromKubeConfig(raw)
```
EItanya: Currently this returns a generic kubernetes error which is not helpful in determining which config is the broken one. Add some context to this error

PR #154, opened 2019-04-26
Crd util

rickducott:


pkg/utils/kubeutils/kube_crd.go
```diff
@@ -0,0 +1,35 @@
...
+		},
+		Spec: v1beta1.CustomResourceDefinitionSpec{
+			Group: skCrd.Group,
+			Names: v1beta1.CustomResourceDefinitionNames{
+				Kind:       skCrd.KindName,
+				ListKind:   skCrd.KindName + "List",
+				Plural:     skCrd.Plural,
+				ShortNames: []string{skCrd.ShortName},
+			},
+			Scope:   "Namespaced",
```
ilackarms: this value can come from the `crd.ClusterScoped` variable

pkg/utils/kubeutils/kube_crd.go
```diff
@@ -0,0 +1,35 @@
+package kubeutils
+
+import (
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/crd"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// The standard CRD we use for helm charts, based on the solo kit CRD
```
mitchdraft: not sure if this comment adds much or will be true in the future
I believe it is describing the return type, but that is already self-documenting

pkg/utils/kubeutils/kube_crd.go
```diff
@@ -0,0 +1,35 @@
...
+func GetCustomResourceDefinition(skCrd crd.Crd, labels map[string]string) *v1beta1.CustomResourceDefinition {
+	return &v1beta1.CustomResourceDefinition{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "CustomResourceDefinition",
+			APIVersion: "apiextensions.k8s.io/v1beta1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: skCrd.FullName(),
+			Annotations: map[string]string{
+				"helm.sh/hook": "crd-install",
```
mitchdraft: might pass this as an argument for greater generality, or else should rename function to reference helm

rickducott: That was the point of the comment..

rickducott: Added to name, agree that's better

PR #157, opened 2019-04-29
namespace by-resource

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/158


pkg/code-generator/codegen/templates/event_loop_test_template.go
```diff
@@ -25,13 +30,22 @@ import (
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients/factory"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients/memory"
+
+	// Needed to run tests in GKE
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+
+	// From https://github.com/kubernetes/client-go/blob/53c7adfd0294caa142d961e1f780f74081d5b15f/examples/out-of-cluster-client-configuration/main.go#L31
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
mitchdraft: duplicate import here

mitchdraft: this gets all cloud providers, fwiw:
`_ "k8s.io/client-go/plugin/pkg/client/auth"`
learned through:
https://github.com/solo-io/squash/commit/5b4aae1b1619f6c2e061906c07e380ade10fa164#diff-70ee2ee3d399e05b88574a8000985f42R11

test/mocks/api/v1/mock_resources.proto
```diff
@@ -49,6 +49,7 @@ message MockResource {
     string data = 1 [json_name="data.json", (gogoproto.jsontag) = "data.json"];
     string some_dumb_field = 100 [(core.solo.io.skip_hashing) = true];
 
+    // comments that go above the basic field in our docs=
```
mitchdraft: might be a stray comment

EItanya: No I believe that comment is there to show how comments in that location wind up in our docs

pkg/code-generator/codegen/templates/event_loop_test_template.go
EItanya: Fixed thay

pkg/api/v1/clients/client_interface.go
```diff
@@ -19,16 +20,46 @@ func DefaultNamespaceIfEmpty(namespace string) string {
 	return namespace
 }
 
+type ResourceWatcher interface {
+	// Deprecated: implemented only by the kubernetes resource client. Will be removed from the interface.
+	Register() error
```
marcogschmidt: I know this is not strictly related to this PR, but could we use the opportunity given by this breaking change to remove the `Register` function from the RC interface? It is a no-op for all RC implementations except for the `kube` one. For the kube client we can move the logic it implements into the constructor.

mitchdraft:
I like this idea, I'm worried about the complexity it adds to configuration.

I think it would help to add some README content that shows how to configure this, both for easier adoption and for communication of the benefits.

There seem to be a lot of diffs unrelated to the change. For example, it's not clear to me why this was deleted
https://github.com/solo-io/solo-kit/pull/157/files#diff-957eee4b05160f500e479b6e83a95419L17

Since this is a breaking change, I'd be more comfortable reviewing the minimal necessary diff.

I see you've got this in draft status, I'd be glad to take another look when you want to merge

EItanya:
Closing in favor of @ilackarms work on multi cluster

Issue #150, opened 2019-04-24
import other project solo-kit.json


EItanya:
in order to include other projects go resources in event loops requires the ability to import other projects solo-kit.json

PR #159, opened 2019-04-30
Mockgen

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/150
resolves https://github.com/solo-io/solo-kit/issues/160


EItanya:
> very nice. 2 comments:
> 
>     * not sure there's any value to mocking out the `Cloneable X Interface` as its only reason for existence is to make it easier to run the `Implement Interface...` action from the IDE. it's not implemented anywhere
> 
>     * it'd be nice to see a test in solokit that makes use of some of this new generated code

@ilackarms about the `Cloneable` interface, I will see if there is an easy way to skip specific resources, but the program just grabs all interfaces from the file, and skipping specific ones may be more trouble than it is worth.
As far as tests, I'll write a few

pkg/code-generator/cmd/main.go
```diff
@@ -26,6 +28,10 @@ import (
 
 type DocsOptions = options.DocsOptions
 
+const (
+	TestMode = "TEST_MODE"
```
ilackarms: unclear name
`SKIP_MOCK_GEN`?

ci/check-code-gen.sh
```diff
@@ -17,7 +17,7 @@ PATH=/workspace/gopath/bin:$PATH
 
 set +e
 
-make generated-code -B  > /dev/null
+make generated-code -B
```
EItanya: This is useful output because it's solo-kit, so seeing errors here can be informative

Issue #161, opened 2019-05-02
multiwatcher


ilackarms:
create the ability to aggregate watches for resources into a single stream of lists on a single `Watch` function

PR #162, opened 2019-05-02
implement multi watcher

ilackarms:
aggregates watches from different clients into a single list
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/161


pkg/api/v1/clients/wrapper/multi_watcher.go
```diff
@@ -0,0 +1,182 @@
...
+	c.sinks[out] = addWatch
+	c.sinksAccess.Unlock()
+
+	go func() {
+		// context is closed, clean up watch resources
+		<-opts.Ctx.Done()
+		c.sinksAccess.Lock()
+		delete(c.sinks, out)
+		c.sinksAccess.Unlock()
+		close(out)
```
yuval-k: this might panic if another goroutine is just about to send something on the out channel. (before it detects the context was cancelled)

yuval-k: you can use a waitgroup to guarantee that no other go routine is using the channel

pkg/api/v1/clients/wrapper/multi_watcher.go
```diff
@@ -0,0 +1,182 @@
...
+		removeWatcher(w)
+	}
+}
+
+// aggregate resources by the channel they were read from
+type resourcesByWatcher map[clients.ResourceWatcher]resources.ResourceList
+
+func (rbw resourcesByWatcher) merge() resources.ResourceList {
+	var merged resources.ResourceList
+	for _, list := range rbw {
```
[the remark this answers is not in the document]
ilackarms: i added the place where it's called to a locked section

pkg/api/v1/clients/wrapper/multi_watcher.go
```diff
@@ -0,0 +1,198 @@
...
+				case err := <-errs:
+					// if the source starts returning errors, remove its list from the snasphot
+					access.Lock()
+					delete(listsByWatcher, watcher)
+					access.Unlock()
+					aggregatedErrs <- err
+					select {
+					case <-ctx.Done():
+						return
+					case out <- listsByWatcher.merge():
```
yuval-k: merge is not locked here

pkg/api/v1/clients/wrapper/multi_watcher.go
```diff
@@ -0,0 +1,198 @@
...
+		go func() {
+			defer cancel()
+			defer sourceWatches.Done()
+
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case err := <-errs:
+					// if the source starts returning errors, remove its list from the snasphot
```
yuval-k: is it possible from a source to recover from the error?
or is the semantics is that error is the last message?

ilackarms: it can recover if we receive a new list, it gets added back to the snapshot

Issue #204, opened 2019-06-26
support multiple versions of a resource


ilackarms:
solo kit needs some jiggling to be able to distinguish different versions of a resource that have the same name and proto package

PR #205, opened 2019-06-26
Support resource multiversion

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/204


changelog/v0.10.0/get-snapshot-from-cache.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: FIX
```
rickducott: BREAKING_CHANGE

PR #213, opened 2019-07-12
Jsonschema

EItanya:


api/v1/solo-kit.proto
```diff
@@ -20,6 +20,8 @@ message Resource {
     bool cluster_scoped = 3;
     // indicates whether documentation generation has to be skipped for the given resource, defaults to false
     bool skip_docs_gen = 4;
+    // indicates whether json schemas should be generated for the given resource, defaults to false
+    bool generate_json_schema = 5;
```
EItanya: Currently unused, just goes by whether solo-kit resource

PR #214, opened 2019-07-16
Resource conversion, one solo-kit.json per project

joekelley:
WIP: Need to do some more cleanup and validation.

This PR adds support for converting resources from one version to another, and redefines a "project" from the de facto "one solo-kit.json file per version" to the preferred "one solo-kit.json file per set of versioned api groups". The latter is a breaking change.
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/215


pkg/code-generator/cmd/main.go
```diff
@@ -122,80 +126,125 @@ func Generate(opts GenerateOptions) error {
...
-			docs, err := docgen.GenerateFiles(project, genDocs)
-			if err != nil {
-				return err
+				outDir := filepath.Join(gopathSrc(), version.VersionConfig.GoPackage)
+				if err := writeCodeFiles(code, outDir); err != nil {
+					return err
+				}
+
+				genDocs = &DocsOptions{}
+				if ag.DocsDir != "" && (genDocs != nil) {
```
EItanya: genDocs will never be nil

pkg/code-generator/cmd/main.go
```diff
@@ -122,80 +126,125 @@ func Generate(opts GenerateOptions) error {
...
+					}
+				}
 			}
 
-			if out, err := exec.Command("goimports", "-w", path).CombinedOutput(); err != nil {
-				return errors.Wrapf(err, "goimports failed: %s", out)
-			}
-		}
+			if ag.ConversionGoPackage != "" {
+				goPackageSegments := strings.Split(ag.ConversionGoPackage, "/")
```
EItanya: `filepath.Base()`

pkg/code-generator/cmd/main.go
```diff
@@ -539,19 +589,44 @@ func importCustomResources(imports []string) ([]model.CustomResourceConfig, erro
...
+				results = append(results, customResources...)
+			}
 		}
-		results = append(results, customResources...)
 	}
 
 	return results, nil
 }
+
+func writeCodeFiles(code code_generator.Files, outDir string) error {
```
EItanya: Thanks for gathering this into a func

pkg/code-generator/codegen/conversion.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		return vi.LessThan(vj)
+	})
+
+	resourceNameToProjects := make(map[string][]*model.Version)
+
+	for index, project := range projects {
+		for _, res := range project.Resources {
+			// only generate files for the resources in our group, otherwise we import
+			if !project.VersionConfig.IsOurProto(res.Filename) && !res.IsCustom {
```
EItanya: Nit: very similar logic is done in a few places, e.g. code-generator. Maybe can be combined into a function

pkg/code-generator/codegen/conversion.go
```diff
@@ -0,0 +1,165 @@
...
+	code_generator "github.com/solo-io/solo-kit/pkg/code-generator"
+	"github.com/solo-io/solo-kit/pkg/code-generator/codegen/templates"
+	"github.com/solo-io/solo-kit/pkg/code-generator/model"
+)
+
+func GenerateConversionFiles(soloKitProject *model.ApiGroup, projects []*model.Version) (code_generator.Files, error) {
+	var files code_generator.Files
+
+	sort.SliceStable(projects, func(i, j int) bool {
+		vi, err := kubeapi.ParseVersion(projects[i].VersionConfig.Version)
```
EItanya: Could be cool to make this sort function a function in `kubeapi` which takes in the 2 versions and returns the function. So a lambda I guess

pkg/code-generator/model/project.go
```diff
@@ -143,40 +171,47 @@ type XDSResource struct {
...
 	}
-	pc.ProjectFile = path
-	if pc.GoPackage == "" {
-		goPkg, err := detectGoPackageForProject(path)
-		if err != nil {
-			return ProjectConfig{}, err
+
+	skp.ProjectFile = path
+	for _, ag := range skp.ApiGroups {
+		goPackageSegments := strings.Split(ag.ResourceGroupGoPackage, "/")
```
EItanya: `filepath.Base()`

pkg/code-generator/model/project.go
```diff
@@ -143,40 +171,47 @@ type XDSResource struct {
...
 
-	Project      *Project
+	Project      *Version
 	ProtoPackage string // eg. gloo.solo.io
 
 	Filename string // the proto file where this resource is contained
 }
 
-func LoadProjectConfig(path string) (ProjectConfig, error) {
+func LoadProjectConfig(path string) (SoloKitProject, error) {
```
EItanya: Any particular reason this isn't returning a pointer?

pkg/code-generator/parser/parser_resource.go
```diff
@@ -59,40 +52,34 @@ func getResource(resources []*model.Resource, project model.ProjectConfig, cfg m
...
-	// default to using the version matching the project itself
-	// only works for this project's resources
-	for _, res := range possibleResources {
-		if res.GoPackage == project.GoPackage {
-			return res, nil
-		}
-	}
-	return nil, errors.Errorf("found %v resources found which match %v, try specifying a version", len(possibleResources), cfg)
+
+	return possibleResources[0], nil
```
EItanya: The above switch statement is no longer necessary as really you are only checking for empty, so you can just do that in an if

test/mocks/conversion/resource_converter.sk.go
```diff
@@ -0,0 +1,143 @@
...
+		return crd.Copy(src, dst)
+	}
+
+	switch t := src.(type) {
+	case *v1.MockResource:
+		return c.convertDown(c.downConverter.FromV1ToV1Alpha1(t), dst)
+	case *v2alpha1.MockResource:
+		return c.convertDown(c.downConverter.FromV2Alpha1ToV1(t), dst)
+	}
+	return errors.New("unrecognized source type, this should never happen")
```
EItanya: Nit: define this as a var somewhere

test/mocks/conversion/resource_converter_test.go
```diff
@@ -0,0 +1,110 @@
+// Code generated by solo-kit. DO NOT EDIT.
+
```
EItanya: solokit build tag

pkg/code-generator/model/project.go
joekelley: nope, just following the existing pattern

test/mocks/conversion/resource_converter.sk.go
joekelley: lmk if you like the templates/errors approach

pkg/code-generator/codegen/conversion.go
joekelley: agreed, but let's hold off till this comes up again.

pkg/api/v1/clients/kube/crd/util.go
```diff
@@ -0,0 +1,17 @@
...
+func Copy(src, dst SoloKitCrd) error {
+	srcBytes, err := protoutils.MarshalBytes(src)
+	if err != nil {
+		return err
+	}
+	err = protoutils.UnmarshalBytes(srcBytes, dst)
+	if err != nil {
+		return err
+	}
+	return nil
```
ilackarms: why are we doing this copy? there's a resources.Clone function

pkg/api/v1/clients/kube/crd/util.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/solo-io/solo-kit/pkg/utils/protoutils"
+)
+
+func Copy(src, dst SoloKitCrd) error {
+	srcBytes, err := protoutils.MarshalBytes(src)
+	if err != nil {
+		return err
+	}
+	err = protoutils.UnmarshalBytes(srcBytes, dst)
```
ilackarms: nit: 
```go
if err := protoutils.UnmarshalBytes(srcBytes, dst); err != nil {
	return err
}
```

pkg/code-generator/cmd/main.go
```diff
@@ -97,9 +97,11 @@ func Generate(opts GenerateOptions) error {
 		if !compileProtos {
 			return false
 		}
-		for _, proj := range projectConfigs {
-			if strings.HasPrefix(protoFile, filepath.Dir(proj.ProjectFile)) {
-				return true
+		for _, skp := range soloKitProjects {
+			for _, vc := range skp.ApiGroup.VersionConfigs {
+				if strings.HasPrefix(protoFile, filepath.Dir(skp.ProjectFile)+"/"+vc.Version) {
```
ilackarms: this means `solo-kit.json` is now expected to live (always) in the parent directory of the version dirs? if so, it might be useful to provide a log line warning or error if the user somehow misses this point

pkg/api/v1/clients/kube/crd/util.go
joekelley: The copy needs to happen in place onto `dst` -- clone returns a new object

pkg/code-generator/cmd/main.go
```diff
@@ -122,79 +124,115 @@ func Generate(opts GenerateOptions) error {
...
-
-		projectConfig.CustomResources = append(projectConfig.CustomResources, importedResources...)
-
-		for _, desc := range descriptors {
-			if filepath.Dir(desc.ProtoFilePath) == filepath.Dir(projectConfig.ProjectFile) {
-				projectConfig.ProjectProtos = append(projectConfig.ProjectProtos, desc.GetName())
+		for _, vc := range skp.ApiGroup.VersionConfigs {
+			vc.CustomResources = append(vc.CustomResources, importedResources...)
+			for _, desc := range descriptors {
+				if filepath.Dir(desc.ProtoFilePath) == filepath.Dir(skp.ProjectFile)+"/"+vc.Version {
```
ilackarms: seems like we have this check in a few places. can we extract to a function?
also, use `filepath.Join` here. even though it's only required on windows. and if you use windows, you're wrong 😏

pkg/code-generator/parser/parser_resource.go
```diff
@@ -34,16 +34,9 @@ type ProtoMessageWrapper struct {
...
 	matches := func(res *model.Resource) bool {
-		if res.Name == cfg.ResourceName &&
-			(res.ProtoPackage == cfg.ResourcePackage || res.GoPackage == cfg.ResourcePackage) {
-			if cfg.ResourceVersion == "" {
-				return true
-			}
-			return cfg.ResourceVersion == res.Version
-		}
-		return false
+		return res.Name == cfg.ResourceName && (res.ProtoPackage == cfg.ResourcePackage || res.GoPackage == cfg.ResourcePackage)
```
joekelley: this check has to be relaxed in order to access resources from other projects

PR #218, opened 2019-07-18
resuscitate  Consul & Vault storage clients

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/217

### DIFF
--- a/pkg/api/v1/clients/vault/resource_client.go
+++ b/pkg/api/v1/clients/vault/resource_client.go
@@ -6,9 +6,9 @@ import (
 	"reflect"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
-	"github.com/ghodss/yaml"
 	"github.com/hashicorp/vault/api"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
 	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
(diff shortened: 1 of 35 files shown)


test/setup/ports.go
```diff
@@ -0,0 +1,7 @@
+package setup
+
+import "github.com/onsi/ginkgo"
+
+func AllocateParallelPort(basePort int) int {
```
EItanya: nice

pkg/utils/protoutils/marshal.go
```diff
@@ -35,6 +35,33 @@ func MarshalBytes(res resources.Resource) ([]byte, error) {
...
+	}
+
+	if protoInto, ok := into.(proto.Message); ok {
+		return jsonpb.Unmarshal(bytes.NewBuffer(jsn), protoInto)
+	}
+	return json.Unmarshal(data, into)
+}
+
+func MarshalYAML(res resources.Resource) ([]byte, error) {
+	jsn, err := func() ([]byte, error) {
```
EItanya: Any particular reason this is in a closure, not a problem, just seems unecessary and makes it a little harder to read

pkg/api/v1/resources/resource_interface.go
```diff
@@ -14,12 +16,17 @@ import (
 
 const delim = " "
 
-type Resource interface {
+type BaseResource interface {
```
EItanya: Why not keep the resource interface and add a `VersionedInterface`

pkg/api/v1/clients/vault/resource_client_test.go
```diff
@@ -21,9 +22,9 @@ var _ = Describe("Base", func() {
 		secrets clients.ResourceClient
 	)
 	BeforeEach(func() {
-		rootKey = "/secret/" + helpers.RandString(4)
+		rootKey = "bark"
```
EItanya: bark?

pkg/api/v1/clients/vault/resource_client.go
```diff
@@ -18,46 +16,31 @@ import (
...
 	if err != nil {
-		return nil, err
+		return nil, false, errors.Wrapf(err, "parsing data response")
 	}
+	// if deletion time set, the secret was deleted
+	deleted := data.Metadata.DeletionTime != ""
+
 	resource := rc.NewResource()
-	return resource, protoutils.UnmarshalBytes(jsn, resource)
+	return resource, deleted, protoutils.UnmarshalMap(data.Data, resource)
```
EItanya: Couple things about this line/function
1) Could you possibly split this line up, so the error from UnmarshalMap is more explicit.
2) Could you just return the `errors.NewNotExistErr(namespace, name)` here if deleted is true rather than returning the boolean.

pkg/utils/protoutils/marshal.go
ilackarms: it's not much more readable the other way, but i can change it if u prefer:
```go 

func MarshalYAML(res resources.Resource) ([]byte, error) {
	var jsn []byte
	if pb, ok := res.(proto.Message); ok {
		buf := &bytes.Buffer{}
		if err := jsonpbMarshaler.Marshal(buf, pb); err != nil {
			return nil, err
		}
		jsn = buf.Bytes()
	} else {
		var err error 
		jsn, err = json.Marshal(res)
		if err != nil {
			return nil, err
		}
	}
	return yaml.JSONToYAML(jsn)
}
```

pkg/api/v1/clients/vault/resource_client_test.go
ilackarms: 🤷‍♂

pkg/api/v1/clients/vault/resource_client.go
ilackarms: i need to know specifically if the resource was deleted here. whether or not that's an error depends where this function is called from (in list we just ignore it)

pkg/utils/protoutils/marshal.go
EItanya: Thanks, I appreciate it

PR #222, opened 2019-07-19
for for all the watches to be called at least once before issuing a sync

yuval-k:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/pull/222

### DIFF
--- a/pkg/code-generator/codegen/templates/snapshot_emitter_template.go
+++ b/pkg/code-generator/codegen/templates/snapshot_emitter_template.go
@@ -138,11 +138,11 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
 {{- if (not .ClusterScoped) }}
 		/* Setup namespaced watch for {{ .Name }} */
 		{
-			upstreams, err := c.{{ lower_camel .Name }}.List(namespace, clients.ListOpts{Ctx: opts.Ctx, Selector: opts.Selector})
+			{{ lower_camel .PluralName }}, err := c.{{ lower_camel .Name }}.List(namespace, clients.ListOpts{Ctx: opts.Ctx, Selector: opts.Selector})
 			if err != nil {
 				return nil, nil, errors.Wrapf(err, "initial {{ .Name }} list")
 			}
-			initial{{ upper_camel .Name }}List = append(initial{{ upper_camel .Name }}List, upstreams...)
+			initial{{ upper_camel .Name }}List = append(initial{{ upper_camel .Name }}List, {{ lower_camel .PluralName }}...)
 		}
 		{{ lower_camel .Name }}NamespacesChan, {{ lower_camel .Name }}Errs, err := c.{{ lower_camel .Name }}.Watch(namespace, opts)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -191,7 +191,38 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
 		originalSnapshot := {{ .GoName }}Snapshot{}
 		currentSnapshot := originalSnapshot.Clone()
 		timer := time.NewTicker(time.Second * 1)
+
+
+		{{- range .Resources}}
+		{{- if not .ClusterScoped }}
+				{{ lower_camel .PluralName }}ByNamespace := make(map[string]{{ .ImportPrefix }}{{ .Name }}List)
```
yuval-k: @ilackarms is it safe to assume that watch never returns a nil list, even if it's empty?

ilackarms: generated clients return an empty list `var x FooList`
clients implementing the interface can return whatever they want

for what is the context?

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -126,13 +126,24 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
...
 {{- end}}
 
+	currentSnapshot := {{ .GoName }}Snapshot{}
+
 	for _, namespace := range watchNamespaces {
 {{- range .Resources}}
 {{- if (not .ClusterScoped) }}
 		/* Setup namespaced watch for {{ .Name }} */
+		{
+			upstreams, err := c.{{ lower_camel .Name }}.List(namespace, clients.ListOpts{Ctx: opts.Ctx, Selector: opts.Selector})
```
ilackarms: `upstreams` -> `{{ lower_camel .PluralName }}`

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -126,13 +126,24 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
...
+
 {{- end}}
 
+	currentSnapshot := {{ .GoName }}Snapshot{}
+
 	for _, namespace := range watchNamespaces {
 {{- range .Resources}}
 {{- if (not .ClusterScoped) }}
 		/* Setup namespaced watch for {{ .Name }} */
+		{
```
ilackarms: why use a closure here?

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
yuval-k: no need as they are in an internal scope

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
yuval-k: so that i can re-use the variable name upstreams

PR #224, opened 2019-07-22
Revert "Filter out resources with when GVK doesn't match that of the client"

joekelley:
Reverts solo-io/solo-kit#219

### DIFF
--- a/pkg/api/v1/clients/kube/resource_client.go
+++ b/pkg/api/v1/clients/kube/resource_client.go
@@ -173,9 +173,6 @@ func (rc *ResourceClient) Read(namespace, name string, opts clients.ReadOpts) (r
 		}
 		return nil, errors.Wrapf(err, "reading resource from kubernetes")
 	}
-	if !rc.matchesClientGVK(*resourceCrd) {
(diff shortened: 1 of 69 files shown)


pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -381,9 +375,11 @@ func (rc *ResourceClient) Watch(namespace string, opts clients.WatchOpts) (<-cha
 	return resourcesChan, errs, nil
 }
 
-// Checks whether the group version kind of the given resource matches that of the client's underlying CRD:
-func (rc *ResourceClient) matchesClientGVK(resource v1.Resource) bool {
-	return resource.GroupVersionKind().String() == rc.crd.GroupVersionKind().String()
+// Checks whether the type of the given resource matches the one of the client's underlying CRD:
+// 1. the kind name must match that of CRD
+// 2. the version must match the CRD GroupVersion (in the form <GROUP_NAME>/<VERSION>)
+func (rc *ResourceClient) matchesClientKind(resource v1.Resource) bool {
```
EItanya: Was this code here before as well?

joekelley: yeah

PR #271, opened 2019-09-13
add ready channel to event loop that is closed after the first sync is called

yuval-k:
part of https://github.com/solo-io/gloo/issues/1191
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/pull/271


changelog/v0.10.18/ready.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
```
EItanya: definitely user facing, I'd consider it a new feature

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -93,12 +93,16 @@ func init() {
 	)
 }
 
+type {{ .GoName }}SnapshotEmitter interface {
+	Snapshots(watchNamespaces []string, opts clients.WatchOpts) (<-chan *{{ .GoName }}Snapshot, <-chan error, error)
+}
+
 type {{ .GoName }}Emitter interface {
+	{{ .GoName }}SnapshotEmitter
```
ilackarms: the naming here is confusing. what's the difference between an emitter and a snapshotemitter?

changelog/v0.10.18/ready.yaml
yuval-k: i guess it depends who is the user. i can change

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -93,12 +93,16 @@ func init() {
 	)
 }
 
+type {{ .GoName }}SnapshotEmitter interface {
```
ilackarms: do we need another interface here?

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
yuval-k: the event loop only needs the Snapshots() function hence the separation of interfaces.
This make the eventloop easier to test. Plus since it doesn't need to full interface, no need to expose it.

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
yuval-k: see comment above

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
ilackarms: ok but `Emitter` and `SnapshotEmitter` are interchangeable in terms of their meaning. can we come up with names that make it clear?

also, i'm not sure it's actually important for us to create a new interface type. since Emitter (as far as i know) is always generated

yuval-k: i want to use the new interface as the event loop argument as the event loop doesn't need the other methods.
what names do you have in mind?

changelog.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
EItanya: Might be cool to add this to our homebrew tap so that we can pull and keep it updated without attaching it to a specific repo, or having to copy paste it around

yuval-k: yea.. i also didn't mean to push this! i'll remove

PR #291, opened 2019-10-08
add helper methods for V2 Reporter

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/288

### DIFF
--- a/pkg/api/v2/reporter/reporter.go
+++ b/pkg/api/v2/reporter/reporter.go
@@ -34,6 +34,12 @@ func (e ResourceReports) Merge(resErrs ResourceReports) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/v2/reporter/reporter.go
```diff
@@ -52,6 +64,15 @@ func (e ResourceReports) AddWarning(res resources.InputResource, warning string)
 	e[res] = rpt
 }
 
+func (e ResourceReports) Find(kind string, ref core.ResourceRef) (resources.InputResource, Report) {
```
marcogschmidt: Looks like this argument is unused.

Issue #303, opened 2019-10-18
race condition in generating proto file descriptors


kdorosh:
found in our CI, sample failure: https://console.cloud.google.com/cloud-build/builds/df18dafe-35cf-4562-adb8-b6e749aab9c2?project=solo-public

sample success: https://console.cloud.google.com/cloud-build/builds/3aec1d7d-114b-4626-b6a1-f007d1af60d6?project=solo-public

We process descriptors in parallel to parallelize io (we create temp files while parsing protos), but we also process these descriptors in the mutex and deduplicate them.

The problem is that the name of the proto doesn't matter toward proto equality, so which proto we use is nondeterministic.

We should deduplicate protos after sorting, outside of the concurrent codepath.

PR #304, opened 2019-10-18
Fix race condition while generating proto descriptors

kdorosh:
Deduplicate proto descriptors after sorting so the name of any duplicated protos we end up using is deterministic
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/303


kdorosh:
Tested against Gloo codegen, could recreate flake after looping ~30 seconds.

After these changes, cannot reproduce after 5-10min+ of looping

Ready for review 👍

pkg/code-generator/cmd/main.go
```diff
@@ -337,20 +337,6 @@ func collectDescriptorsFromRoot(root string, customImports, customGogoArgs, skip
...
-			existingCopy := proto.Clone(existing.FileDescriptorProto).(*descriptor.FileDescriptorProto)
-			existingCopy.Name = f.Name
-			if proto.Equal(existingCopy, f.FileDescriptorProto) {
-				// if this proto file first came in as an import, but later as a solo-kit project proto,
-				// ensure the original proto gets updated with the correct proto file path
-				if existing.ProtoFilePath == "" {
-					existing.ProtoFilePath = f.ProtoFilePath
-				}
-				return
-			}
```
kdorosh: for anyone reading, the problem is that this logic is within the async code. descriptors can be equal if their names aren't the same, thus the first descriptor found will win out.

by moving this logic out of the async code (and after we sort the descriptors using their name), we can get deterministic results

PR #307, opened 2019-10-21
Less reflection

yuval-k:


pkg/api/v1/clients/memory/resource_client_test.go
```diff
@@ -57,6 +60,35 @@ var _ = Describe("Base", func() {
 		listret2, err := client.List("ns", clients.ListOpts{})
 		Expect(err).NotTo(HaveOccurred())
 		Expect(listret[0]).NotTo(BeIdenticalTo(listret2[0]))
+	})
+
+	Context("Benchmarks", func() {
+		Measure("it should do something hard efficiently", func(b Benchmarker) {
```
ilackarms: do you think we can write this Measure with something more meaningful?

changelog/v0.11.4/less_reflection.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Reduce usage of reflection
```
ilackarms: for the changelog, would you mind enumerating the individual changes?

PR #310, opened 2019-10-25
coalesce kube udpates to 1 a second. 

yuval-k:
they are slow as they need to convert each object to its sk type


pkg/api/v1/clients/common/common.go
```diff
@@ -58,14 +60,23 @@ func KubeResourceWatch(cache cache.Cache, listFunc ResourceListFunc, namespace s
 		defer close(errs)
 
 		// watch should open up with an initial read
+		timer := time.NewTicker(time.Second)
```
ilackarms: can we make this configurable? 1s seems kind of slow for some use cases

also, what do you think of a feature flag so it can be disabled? (in the kube factory)

yuval-k: i can't imagine 1s to be slow for any use case... what use case do you have in mind?
same goes for disabling this - i'm not sure i see a clear benefit that mertis this.. can you elaborate more?

ilackarms: on second thought i think it's ok for solo-kit considering this function is only used (practically) in event loops.

Issue #312, opened 2019-10-29
Populate cluster name on kube meta


joekelley:
We need to populate the cluster name field on kube meta when we are converting from solo-kit metadata to kube metadata.

PR #313, opened 2019-10-29
populate clusterName field on kubernetes resources

joekelley:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/312

### DIFF
--- a/changelog/v0.11.8/kube-meta-cluster.yaml
+++ b/changelog/v0.11.8/kube-meta-cluster.yaml
@@ -1,0 +1,4 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Add clusterName to metadata on kube objects.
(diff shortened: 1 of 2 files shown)


joekelley:
It'll allow us to know what cluster the resources came on read

ilackarms:
clusterName will not be persisted on write - is that OK ?

joekelley:
Yep! This is just to mimic the behavior we get for kube crd clients where the name of the cluster a resource is read from can be found on the solo-kit metadata fields on the resource.

I understand that it feels strange to use this field that the k8s apiserver explicitly ignores, but this feels like a valid use case to me. Persisting cluster name wouldn't be very useful anyway, since at this point cluster name only represents the name a user provides in a kube config secret and has no meaning outside that context.

PR #323, opened 2019-11-26
Bump go-utils

marcogschmidt:
Bump the `go-utils` dependency to `0.11.0`, which contains libraries compatible with Helm 3. This bump requires the update of the kubernetes libraries to `1.16.x` to work.


pkg/api/v1/clients/kube/crd/client/clientset/versioned/scheme/register.go
```diff
@@ -20,31 +20,36 @@ package scheme
...
+////     aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
+////   )
+////
+////   kclientset, _ := kubernetes.NewForConfig(c)
+////   _ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
+////
+//// After this, RawExtensions in Kubernetes types will serialize kube-aggregator types
+//// correctly.
+//var AddToScheme = localSchemeBuilder.AddToScheme
+
```
ilackarms: nit ^

marcogschmidt: Will remove

Issue #326, opened 2019-12-12
migrate to go mod


EItanya:
migrate proto generation/vendoring to go mod.

PR #328, opened 2019-12-16
Go.mod

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/326


pkg/api/v1/clients/file/resource_client_test.go
```diff
@@ -22,13 +22,13 @@ var _ = Describe("Base", func() {
 		client *ResourceClient
 		tmpDir string
 	)
-	BeforeEach(func() {
+	JustBeforeEach(func() {
```
yuval-k: why is this needed?

pkg/code-generator/codegen/generator.go
```diff
@@ -141,7 +141,7 @@ func generateFilesForResourceGroup(rg *model.ResourceGroup) (code_generator.File
 	} {
 		content, err := generateResourceGroupFile(rg, tmpl)
 		if err != nil {
-			return nil, errors.Wrapf(err, "internal error: processing %template '%v' for resource group %v failed", tmpl.ParseName, rg.Name)
+			return nil, errors.Wrapf(err, "internal error: processing %template '%s' for resource group %v failed", tmpl.ParseName, rg.Name)
```
yuval-k: what's the "%t"?

EItanya: Hahaha
That’s been there’s while I guess
Probably a mistake

pkg/api/v1/clients/file/resource_client_test.go
EItanya: It was failing without it

PR #346, opened 2020-02-11
clean up external envoy API docs

ashleywang1:
When we render our external API docs from external envoy protos, the comments get rendered incorrectly. Envoy uses some formatting in their comments that we don't use, (e.g. :ref:``).
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/2291


ashleywang1:
This needs to be merged first before https://github.com/solo-io/gloo/pull/2402

PR #351, opened 2020-03-20
Fix overactive watch

kdorosh:
WIP


pkg/api/v1/clients/kube/controller/controller.go
```diff
@@ -187,9 +197,36 @@ func (c *Controller) enqueueSync(t eventType, old, new interface{}) {
 		runtime.HandleError(err)
 		return
 	}
+	// if no relevant changes were made, no reason to queue up work
+	if reflect.DeepEqual(old, new) {
```
marcogschmidt: Could this be a problem with very large config maps? The fact that we were hashing the `data` was causing huge CPU spikes, does `deepEquals` would compare those fields as well?

pkg/api/v1/clients/kube/controller/controller.go
```diff
@@ -187,9 +197,36 @@ func (c *Controller) enqueueSync(t eventType, old, new interface{}) {
...
+
+	oldYaml := SprintAny(old)
+	newYaml := SprintAny(new)
+	if oldYaml == newYaml {
+		return // reflect deep equals appears to work here, may want to remove this
+	}
+
+	// being more safe than we have to be.. remove??
+	// we should ensure event is namespaced to our watch namespaces!
+	if strings.Contains(key, "kube-system") {
```
marcogschmidt: That's a very good idea! We could derive the namespaces we care about from the `namespaceWhitelist` on the `resourceClients` that we register with the `ResourceClientSharedInformerFactory`.

pkg/api/v1/clients/kube/controller/controller.go
kdorosh: Perhaps, if we can afford to drop events when the buffer gets full that may be a better solution. this requires more thought & testing

pkg/api/v1/clients/kube/controller/controller.go
kdorosh: i believe the informer registry is already namespaced, so i might have been able to get away with changing the `watchNamespaces` here

kdorosh:
superseded by https://github.com/solo-io/solo-kit/pull/353

PR #352, opened 2020-03-24
Change to key so we have single event per resource on workqueue

kdorosh:


kdorosh:
superseded by https://github.com/solo-io/solo-kit/pull/353

PR #356, opened 2020-04-29
Fix race condition

ashleywang1:


pkg/api/v2/reporter/reporter.go
```diff
@@ -165,11 +165,15 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 }
 
 func attemptUpdateStatus(ctx context.Context, client clients.ResourceClient, resourceToWrite resources.InputResource) (resources.Resource, resources.InputResource, error) {
+	var readErr error
+	_, readErr = client.Read(resourceToWrite.GetMetadata().Namespace, resourceToWrite.GetMetadata().Name, clients.ReadOpts{Ctx: ctx})
+	if readErr != nil { // resource has been deleted, don't re-create
+		return resourceToWrite, resourceToWrite, nil
```
kdorosh, line 171: should be
```suggestion
		return nil, resourceToWrite, nil
```
as the return args are `// updatedResource, resourceToWrite, error`

kdorosh: and then we need to wrap
```golang
		resources.UpdateMetadata(resource, func(meta *core.Metadata) {
			meta.ResourceVersion = updatedResource.GetMetadata().ResourceVersion
		})

		logger.Debugf("wrote report %v : %v", updatedResource.GetMetadata().Ref(), status)
```
in a nil check on `updatedResource`

pkg/api/v2/reporter/reporter.go
```diff
@@ -155,21 +155,27 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
...
-		resources.UpdateMetadata(resource, func(meta *core.Metadata) {
-			meta.ResourceVersion = updatedResource.GetMetadata().ResourceVersion
-		})
+		if updatedResource != nil {
+			resources.UpdateMetadata(resource, func(meta *core.Metadata) {
+				meta.ResourceVersion = updatedResource.GetMetadata().ResourceVersion
+			})
+		}
 
 		logger.Debugf("wrote report %v : %v", updatedResource.GetMetadata().Ref(), status)
```
kdorosh, line 164: this also should be in the nil check. if the updated resource is nil, we didn't write a report. (and should log that we didn't write a report)

kdorosh: we also should look into setting the resource that no longer exists to nil in `ResourceReports`, as we were side affecting those when we did successfully update the status.

this could be a dangerous change, need to look at the places we use `WriteReports` to make sure that it's safe

ashleywang1: I checked both the gloo and gateway syncers - both don't use the ResourceReports reports after it's passed in to the WriteReports call. When the new snapshot comes in and gets validated, which creates the new report, it should be updated to no longer have the resource that no longer exists.

pkg/api/v2/reporter/reporter_test.go
```diff
@@ -97,6 +97,18 @@ var _ = Describe("Reporter", func() {
...
+		It("checks to make sure a resource exists before writing to it", func() {
+			res := v1.NewMockResource("", "mocky")
+			resourceErrs := rep.ResourceReports{
+				res: rep.Report{Errors: fmt.Errorf("pocky")},
+			}
+
+			mockedResourceClient.EXPECT().Read(res.Metadata.Namespace, res.Metadata.Name, gomock.Any()).Return(nil, errors.NewNotExistErr("", "mocky"))
+
+			err := reporter.WriteReports(context.TODO(), resourceErrs, nil)
+			Expect(err).NotTo(HaveOccurred())
```
kdorosh, line 109: can we actually assert that the `mockedResourceClient.Write()` call was never made?

can we assert that `resourceErrs` is an empty map now (will need to update the logic to remove the entry from the map)?

pkg/api/v2/reporter/reporter.go
```diff
@@ -155,21 +155,28 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
...
-		})
-
-		logger.Debugf("wrote report %v : %v", updatedResource.GetMetadata().Ref(), status)
+		if updatedResource != nil {
+			resources.UpdateMetadata(resource, func(meta *core.Metadata) {
+				meta.ResourceVersion = updatedResource.GetMetadata().ResourceVersion
+			})
+			logger.Debugf("wrote report for %v : %v", updatedResource.GetMetadata().Ref(), status)
+		} else {
+			logger.Debugf("did not write report for %v : %v because resource was not found", resourceToWrite.GetMetadata().Ref(), status)
```
kdorosh, line 164: in this case we should make sure that we remove `resource` from the `ResourceReports` map. to do this safely, we will need to copy the map before iterating over the copy, so we can remove from the original map

pkg/api/v2/reporter/reporter.go
```diff
@@ -155,21 +155,28 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 			merr = multierror.Append(merr, err)
```
kdorosh, line 155: it would also be prudent to add a comment to this function that mentions that `ResourceReports` will be modified with the results after writing the reports, noting that there may be fewer resources that originally requested if they didn't exist

PR #357, opened 2020-05-07
don't Find resources without a namespace

arianaw66:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/358


arianaw66:
Got into a bad state, replaced with https://github.com/solo-io/solo-kit/pull/359

PR #360, opened 2020-06-01
make WriteReports safer

arianaw66:
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/2954

### DIFF
--- a/pkg/api/v2/reporter/reporter.go
+++ b/pkg/api/v2/reporter/reporter.go
@@ -177,11 +177,22 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 	return merr.ErrorOrNil()
 }
 
+// Ideally, this and its caller, WriteReports, would just take the resource ref and its status, rather than the resource itself,
+//    to avoid confusion about whether this may update the resource rather than just its status.
+//    However, this change is not worth the effort and risk right now. (Ariana, June 2020)
 func attemptUpdateStatus(ctx context.Context, client clients.ResourceClient, resourceToWrite resources.InputResource) (resources.Resource, resources.InputResource, error) {
 	var readErr error
-	_, readErr = client.Read(resourceToWrite.GetMetadata().Namespace, resourceToWrite.GetMetadata().Name, clients.ReadOpts{Ctx: ctx})
+	resourceFromRead, readErr := client.Read(resourceToWrite.GetMetadata().Namespace, resourceToWrite.GetMetadata().Name, clients.ReadOpts{Ctx: ctx})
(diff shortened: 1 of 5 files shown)


pkg/api/v2/reporter/reporter.go
```diff
@@ -177,12 +177,23 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
...
 func attemptUpdateStatus(ctx context.Context, client clients.ResourceClient, resourceToWrite resources.InputResource) (resources.Resource, resources.InputResource, error) {
 	var readErr error
-	_, readErr = client.Read(resourceToWrite.GetMetadata().Namespace, resourceToWrite.GetMetadata().Name, clients.ReadOpts{Ctx: ctx})
+	resourceFromRead, readErr := client.Read(resourceToWrite.GetMetadata().Namespace, resourceToWrite.GetMetadata().Name, clients.ReadOpts{Ctx: ctx})
 	if readErr != nil && errors.IsNotExist(readErr) { // resource has been deleted, don't re-create
 		return nil, resourceToWrite, nil
 	}
+	if readErr == nil {
+		status := resourceToWrite.GetStatus()
+		// set resourceToWrite to the resource we read but with the new status
```
kdorosh, line 191: perhaps note in a comment that it's possible that the resource read is more recent than the one we calculated status for, but that this is fine because we will soon recalculate the status and update again.

it's not a big deal if the status is temporarily (fractions of a second) wrong for the resource because the status is should not be acted on, and is informational only. further, the status is accurate for the resource as it's stored in gloo's memory for that timeframe

the alternatives are:
- (original behavior) wipe the entire resource with the resource as calculated, which picks up in-memory modifications to the resource that we don't want to write (and that can confuse users and can introduce future bugs as gloo writes config the user didn't provide)
- wipe the entire resource with the original resource as originally calculated from the snapshot, but this doubles the size of snapshot use in memory and the alternative (status _possibly_ wrong harmlessly for fractions of a second) seems better

pkg/api/v2/reporter/reporter.go
```diff
@@ -177,12 +177,23 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 	return merr.ErrorOrNil()
 }
 
+// Ideally, this and its caller, WriteReports, would just take the resource ref and its status, rather than the resource itself,
+//    to avoid confusion about whether this may update the resource rather than just its status.
+//    However, this change is not worth the effort and risk right now. (Ariana, June 2020)
```
kdorosh, line 182: 👏

PR #362, opened 2020-07-07
Make resource status type generic

marcogschmidt:
Change the type of the status field of the core `Resource` type to a generic `map[string]interface{}` in order to be able to handle resources with any status format.


EItanya:
🎉

pkg/utils/protoutils/marshal.go
```diff
@@ -98,6 +99,19 @@ func MarshalMapEmitZeroValues(from resources.Resource) (map[string]interface{},
 	return m, err
 }
 
+func MarshalMapFromProto(from proto.Message) (map[string]interface{}, error) {
+	out := &bytes.Buffer{}
+	if err := jsonpbMarshaler.Marshal(out, from); err != nil {
```
EItanya, line 104: If we are worried about backwards compatibility issues here, it may be worth setting `EnumsAsInts` on this marshaller as well. In gloo-fed we ran into a bug where sk clients couldn't read enums as strings because of how the unmarshalling worked. If we are worried about having both in cluster at the same time during upgrades, or other situations, we should most likely do this.

pkg/utils/protoutils/marshal.go
```diff
@@ -98,6 +99,19 @@ func MarshalMapEmitZeroValues(from resources.Resource) (map[string]interface{},
...
+		panic(fmt.Sprintf("failed to marshal proto to bytes: %v", err))
```
EItanya, line 105: Why are we panicking here and not just returning the error and leaving it up to the caller. If this is intended we should remove the `error` return value and call this function `MustMarshalMapFromProto`

pkg/utils/protoutils/marshal.go
```diff
@@ -98,6 +99,19 @@ func MarshalMapEmitZeroValues(from resources.Resource) (map[string]interface{},
...
+	}
+
+	var m map[string]interface{}
+	if err := json.Unmarshal(out.Bytes(), &m); err != nil {
+		panic(fmt.Sprintf("failed to unmarshal bytes to map: %v", err))
```
EItanya, line 110: Same as above

pkg/utils/protoutils/marshal.go
marcogschmidt: Oh good point, I missed this when moving the logic into this util function. Will fix.

pkg/utils/protoutils/marshal.go
marcogschmidt: Could you give more details about the problem you encountered in gloo-fed? I extensively tested this change in Gloo, so I am curious about what corner case I might have missed. I am reluctant to change these marshalling utils unless it is absolutely necessary.

marcogschmidt: Talked to @EItanya , reporting the details of the conversation here for posterity. The issue is that if we were to use the current marshaller for the status, we would write the `status.state` enums as strings. This would work fine with the new solo kit, as we use the proto unmarshaller to deserialize the string enums into the `status` proto message. The problem arises, like Eitan mentioned above, when you consider what happens when older versions of solo-kit try to read in the new string enums: the older version let the k8s `client-go` library deserialize the status into the `status` proto using the default Go JSON unmarshaller, which would fail when the enums are represented as strings because the Go type for enums is `int32`. 

Since it is conceivable that we will have both old and new versions of solo-kit running in the same cluster, I will change the function to use a dedicated marshaller with `EnumsAsInts` set to true.

pkg/api/v1/clients/kube/crd/crd.go
```diff
@@ -74,17 +74,29 @@ func (d Crd) Register(apiexts apiexts.Interface) error {
...
+	// Handle spec
 	data, err := protoutils.MarshalMap(resource)
 	if err != nil {
 		panic(fmt.Sprintf("internal error: failed to marshal resource to map: %v", err))
 	}
+
 	delete(data, "metadata")
 	delete(data, "status")
 	spec := v1.Spec(data)
+
```
EItanya, line 87: Can't you potentially just do the same here with the status, rather than marshalling again?
```
status, ok := data["status"]
if ok {
    status := v1.Status(status)
}
```

marcogschmidt: Potentially, but I think we should use the `GetStatus()` function on the input `InputResource` as the source of truth for the status.

PR #364, opened 2020-07-09
Custom spec/status conversions

marcogschmidt:
Allow custom resource to define their own marshalling behavior for `spec` and `status` fields.

### DIFF
--- a/pkg/api/v1/clients/kube/resource_client.go
+++ b/pkg/api/v1/clients/kube/resource_client.go
@@ -198,7 +198,10 @@ func (rc *ResourceClient) Write(resource resources.Resource, opts clients.WriteO
 	// mutate and return clone
 	clone := resources.Clone(resource).(resources.InputResource)
 	clone.SetMetadata(meta)
-	resourceCrd := rc.crd.KubeResource(clone)
+	resourceCrd, err := rc.crd.KubeResource(clone)
+	if err != nil {
+		return nil, err
+	}
 
 	ctx := opts.Ctx
 	if ctxWithTags, err := tag.New(ctx, tag.Insert(KeyKind, rc.resourceName), tag.Insert(KeyOpKind, "write")); err == nil {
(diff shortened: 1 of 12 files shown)


pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -423,22 +423,40 @@ func (rc *ResourceClient) convertCrdToResource(resourceCrd *v1.Resource) (resour
...
+				return nil, errors.Wrapf(err, "unmarshalling crd spec on custom resource %v in namespace %v into %v",
+					resourceCrd.Name, resourceCrd.Namespace, rc.resourceName)
 			}
-			*status = typedStatus
-			return nil
 		}
-		if err := resources.UpdateStatus(withStatus, updateFunc); err != nil {
-			return nil, err
+		if err := customResource.UnmarshalStatus(resourceCrd.Status); err != nil {
+			return nil, errors.Wrapf(err, "unmarshalling crd status on custom resource %v in namespace %v into %v",
```
EItanya, line 436: I'm worried about this as the underlying implementation will panic every time

pkg/api/v2/reporter/reporter.go
```diff
@@ -103,18 +103,25 @@ func (e ResourceReports) ValidateStrict() error {
 	return errs
 }
 
+// Minimal set of client operations required for reporters.
+type ReporterResourceClient interface {
+	Kind() string
+	Read(namespace, name string, opts clients.ReadOpts) (resources.Resource, error)
+	Write(resource resources.Resource, opts clients.WriteOpts) (resources.Resource, error)
+}
```
EItanya, line 111: All lgtm, but I don't have any context on this change

marcogschmidt: The interface is the subset of the functions of the `ResourceClient` interface that are actually used by the reporter.

In Gloo, I will provide a special implementation for the `RateLimitConfig` that uses the `Update*Status` function on the skv2 client to update the status sub-resource (the normal `Update` performed by the internal sk clientset would not work on subresources). With this interface I'll just need to implement the functions that are really needed (no `List`, `Delete`, etc.).

marcogschmidt:
The libraries are not used anywhere and we deleted the cluster where the tests were running.

PR #369, opened 2020-08-10
Healthcheck basicauth

jehawley:
Add ability to specify extra headers as secrets. Change tested with downstream implementation in gloo.


changelog/v0.13.11/healthcheck-secret-header.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    description: Allow users to specify extra headers as secrets.
+    issueLink: https://github.com/solo-io/gloo/issues/29146
```
kdorosh, line 4: broken link

api/external/envoy/api/v2/core/base.proto
```diff
@@ -180,10 +180,22 @@ message HeaderValue {
...
 // Header name/value pair plus option to control append behavior.
 message HeaderValueOption {
   // Header name/value pair that this option applies to.
-  HeaderValue header = 1 [(validate.rules).message = {required: true}];
+  oneof header_option {
+    option (validate.required) = true;
+    // Explicitly specified herader
+    HeaderValue header = 1;
+    // Reference to header contained in a secret
+    HeaderSecretValue header_secret_value = 3;
```
kdorosh, line 197: let's call this `header_secret_ref`

api/external/envoy/api/v2/core/base.proto
```diff
@@ -180,10 +180,22 @@ message HeaderValue {
...
+
 // Header name/value pair plus option to control append behavior.
 message HeaderValueOption {
   // Header name/value pair that this option applies to.
-  HeaderValue header = 1 [(validate.rules).message = {required: true}];
+  oneof header_option {
+    option (validate.required) = true;
+    // Explicitly specified herader
+    HeaderValue header = 1;
+    // Reference to header contained in a secret
```
kdorosh, line 196: let's include some docs here that explain how the secret will get used (key/values in secret becoming the header key/values)

jehawley: I don't think that's appropriate here? solo-kit isn't responsible for how these get used downstream.

changelog/v0.13.11/healthcheck-secret-header.yaml
jehawley: Fixed.

api/external/envoy/api/v2/core/base.proto
jehawley: Done.

kdorosh: did you forget to push a commit?

api/external/envoy/api/v2/core/base.proto
kdorosh: These docs get rendered into the gloo docs https://docs.solo.io/gloo/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/headers/headers.proto.sk/#headervalueoption

and no other project uses solo-kit (or currently plans to) beyond Gloo, so I think a bit more detail is merited here. 

i would propose a bit more detail, but not including anything gloo-specific (avoiding the word gloo/any of its implementation details, e.g. translation loop), but instead focus on the secret format and how it will get parsed. thoughts

jehawley: In principle, we could radically change how we implement this gloo-side and it would occur to no one to update this doc. This secret ref is totally independent of the secret shape. And the only secret with the correct shape on the gloo side is the HeaderSecret; referencing that here seems like it breaks the "don't reference gloo" principle.

api/external/envoy/api/v2/core/base.proto
```diff
@@ -180,10 +180,22 @@ message HeaderValue {
...
+    string namespace = 2;
+}
+
 // Header name/value pair plus option to control append behavior.
 message HeaderValueOption {
   // Header name/value pair that this option applies to.
-  HeaderValue header = 1 [(validate.rules).message = {required: true}];
+  oneof header_option {
+    option (validate.required) = true;
+    // Explicitly specified herader
```
kdorosh, line 194: typo,

```suggestion
    // Explicitly specified header
```

api/external/envoy/api/v2/core/base.proto
```diff
@@ -180,10 +180,22 @@ message HeaderValue {
...
 // Header name/value pair plus option to control append behavior.
 message HeaderValueOption {
   // Header name/value pair that this option applies to.
-  HeaderValue header = 1 [(validate.rules).message = {required: true}];
+  oneof header_option {
+    option (validate.required) = true;
+    // Explicitly specified header
+    HeaderValue header = 1;
+    // Reference to header contained in a secret
+    HeaderSecretRef header_secret_ref = 3;
```
kdorosh, line 197: We should change this to `ResourceRef`

kdorosh: https://github.com/solo-io/solo-kit/blob/master/api/v1/ref.proto

PR #370, opened 2020-08-11
Bump k8s 1.18.x

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/371


Makefile
```diff
@@ -31,16 +31,21 @@ mod-download:
...
-	GO111MODULE=off go get -u github.com/gogo/protobuf/protoc-gen-gogo
-	GO111MODULE=off go get -u github.com/golang/protobuf/protoc-gen-go
-	GO111MODULE=off go get -u github.com/envoyproxy/protoc-gen-validate
-	GO111MODULE=off go get -u github.com/golang/mock/gomock
-	GO111MODULE=off go install github.com/golang/mock/mockgen
+	GOBIN=$(DEPSGOBIN) go install github.com/solo-io/protoc-gen-ext
+	GOBIN=$(DEPSGOBIN) go install golang.org/x/tools/cmd/goimports
+	GOBIN=$(DEPSGOBIN) go install github.com/gogo/protobuf/protoc-gen-gogo
+	GOBIN=$(DEPSGOBIN) go get -u github.com/golang/protobuf/protoc-gen-go
+	GOBIN=$(DEPSGOBIN) go get -u github.com/envoyproxy/protoc-gen-validate
```
kdorosh, line 44: why `go get` instead of `go install`? feels like we should only be doing `go install` here, or nothing at all

PR #389, opened 2020-11-17
Ripout gogo

EItanya:


api/external/envoy/api/v2/core/base.proto
```diff
@@ -1,27 +1,25 @@
 syntax = "proto3";
 
-package envoy.api.v2.core;
+package solo.io.envoy.api.v2.core;
```
kdorosh, line 3: why add the `solo.io.` prefix everywhere?

go.mod
```diff
@@ -75,7 +73,7 @@ replace (
 	github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309
 
 	// Breaking golang protobuf change
-	github.com/golang/protobuf => github.com/golang/protobuf v1.3.5
+	//github.com/golang/protobuf => github.com/golang/protobuf v1.3.5
```
kdorosh, line 76: remove?

kdorosh: also 🎉

pkg/code-generator/codegen/templates/funcs.go
```diff
@@ -10,10 +10,10 @@ import (
 
 	htmltemplate "html/template"
 
-	"github.com/gogo/protobuf/protoc-gen-gogo/descriptor"
+	"github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/iancoleman/strcase"
 	gendoc "github.com/ilackarms/protoc-gen-doc"
-	"github.com/ilackarms/protokit"
+	"github.com/pseudomuto/protokit"
```
kdorosh, line 16: why this change?

pkg/utils/kubeutils/metadata.go
```diff
@@ -20,13 +20,16 @@ func FromKubeMeta(meta metav1.ObjectMeta) core.Metadata {
...
 	skMeta := ToKubeMetaMaintainNamespace(meta)
 	skMeta.Namespace = clients.DefaultNamespaceIfEmpty(meta.Namespace)
 	return skMeta
 }
 
-func ToKubeMetaMaintainNamespace(meta core.Metadata) metav1.ObjectMeta {
+func ToKubeMetaMaintainNamespace(meta *core.Metadata) metav1.ObjectMeta {
+	if meta == nil {
+		return metav1.ObjectMeta{}
+	}
```
kdorosh, line 32: is this really preferable to returning nil?

test/mocks/tests/generated_kube_types_test.go
```diff
@@ -18,7 +18,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-//go:generate bash ../v2alpha1/kube/hack/update-codegen.sh
+// //go:generate bash ../v2alpha1/kube/hack/update-codegen.sh
```
kdorosh, line 21: clean up?

test/mocks/tests/generated_kube_types_test.go
```diff
@@ -67,8 +67,8 @@ var _ = Describe("Generated Kube Code", func() {
 
 		out, err := testClient.MockResources(res.Namespace).Create(ctx, res, v1.CreateOptions{})
 		Expect(err).NotTo(HaveOccurred())
-		out.Spec.Metadata = core.Metadata{}
-		Expect(out.Spec).To(Equal(res.Spec))
+		// out.Spec.Metadata = &core.Metadata{}
+		Expect(&out.Spec).To(matchers.MatchProto(&res.Spec))
```
kdorosh, line 71: clean up?

kdorosh (commented):
couple questions, but looks generally sane/safe to me

pkg/utils/kubeutils/metadata.go
EItanya: That could have happened before as well. The issue here is that the `metav1.ObjectMeta{}` is never a pointer in kube objects, so this is closer to how it will be used 99% of the time.

pkg/code-generator/codegen/templates/funcs.go
EItanya: The `ilackarms` version was for `gogo`

api/external/envoy/api/v2/core/base.proto
EItanya: This is an unfortunate result of the newer versions of `golang/protobuf`. Essentially if a type is registered twice, it logs a very verbose warning, and will start to panic in an upcoming release. A type is registered by it's `package` + `TypeName`. So these types were generated by us and envoy with the same name, causing them to be "registered twice". In order to solve this I prepended a `solo.io.` to all of our copied/owned envoy protos.

api/external/envoy/api/v2/core/base.proto
```diff
@@ -1,27 +1,25 @@
...
-
-import "gogoproto/gogo.proto";
-option (gogoproto.equal_all) = true;
 
 import "extproto/ext.proto";
 option (extproto.hash_all) = true;
+option (extproto.equal_all) = true;
 
-import "solo-kit/api/v1/ref.proto";
+import "github.com/solo-io/solo-kit/api/v1/ref.proto";
```
yuval-k, line 22: do we have to use the github prefix? this binds us to use go path, no?

pkg/api/v1/apiserver/generate.sh
```diff
@@ -0,0 +1,45 @@
...
+trap "cleanup" EXIT SIGINT
+
+echo ">> Temporary output directory ${TEMP_DIR}"
+
+IMPORTS="\
+    -I=${IN} \
+    -I=${ROOT} \
+    -I=${VENDOR_ROOT}/solo-io/protoc-gen-ext"
+
+GOGO_FLAG="--go_out=plugins=grpc:${TEMP_DIR}"
```
yuval-k, line 33: nit: GO_FLAG

api/external/envoy/api/v2/core/base.proto
yuval-k: discussed on phone

pkg/api/v1/apiserver/generate.sh
yuval-k: also, what is this file?

pkg/code-generator/cmd/main.go
```diff
@@ -238,6 +238,9 @@ func (r *Runner) Run() error {
 
 	var customCompilePrefixes []string
 	for _, relativePath := range r.Opts.CustomCompileProtos {
+		if !strings.HasPrefix(relativePath, anyvendor.DefaultDepDir) {
```
yuval-k, line 241: out of curiosity, what does this change do?

pkg/code-generator/codegen/templates/kube/types_template.go
```diff
@@ -81,14 +81,18 @@ func (o *{{ .Name }}) UnmarshalJSON(data []byte) error {
 	if err := protoutils.UnmarshalResource(data, &spec); err != nil {
 		return err
 	}
+	spec.Metadata = nil
```
yuval-k, line 84: why are we nilling metadata and status below?

pkg/utils/protoutils/marshal.go
```diff
@@ -143,6 +140,18 @@ func UnmarshalMapToProto(m map[string]interface{}, into proto.Message) error {
 	return jsonpb.Unmarshal(bytes.NewBuffer(data), into)
 }
 
+func UnmarshalStruct(structuredData *structpb.Struct, into interface{}) error {
```
yuval-k, line 143: function looks fine - why do we need it here? i dont see it used in the PR?

pkg/api/v1/apiserver/generate.sh
EItanya: Generates the proto for the apiserver client. It's one of the solo-kit backends, I believe it's unused.

pkg/utils/protoutils/marshal.go
EItanya: It's used in Gloo. Essentially the go-utils protoutils will be unusable by gloo after this since they are all gogo proto. But other projects still using gogo will need those protoutils to stay the same. So I'm just updating these to have the go proto versions

EItanya: Going to remove though, because it also exists in gloo already

pkg/code-generator/cmd/main.go
EItanya: There is an option in solo-kit to add custom protos to be compiled along by protoc. The compiler requires them to have the anyvendor prefix, but they are often not passed in with that prefix. So this code adds it just in case.

yuval-k (approved):

PR #394, opened 2020-12-01
Deeper merge

Sodman:


pkg/api/v2/reporter/reporter.go
```diff
@@ -30,9 +30,63 @@ func (e ResourceReports) Accept(res ...resources.InputResource) ResourceReports
...
+				}
+			}
+
+			if firstReport.Errors == nil {
+				// Only 2nd has errs
+				firstReport.Errors = secondReport.Errors
+				e[k] = firstReport
+				continue
+			} else if secondReport.Errors == nil {
+				// Only 1st has errs
```
kdorosh, line 62: don't we still need `e[k] = firstReport` here, in the event we had warnings but no errors in the second report?

pkg/api/v2/reporter/reporter.go
```diff
@@ -30,9 +30,63 @@ func (e ResourceReports) Accept(res ...resources.InputResource) ResourceReports
...
+						allErrsMap[err.Error()] = err
+					}
+					for _, err := range errs2.Errors {
+						if _, found := allErrsMap[err.Error()]; !found {
+							allErrsMap[err.Error()] = err
+							errs1.Errors = append(errs1.Errors, err)
+						}
+					}
+				}
+				firstReport.Errors = errs1
```
kdorosh, line 82: what if `errs1` is multi but `errs2` is not?

kdorosh: and vice-versa

pkg/api/v2/reporter/reporter.go
Sodman: Great catch, added this + added a test for this case!

pkg/api/v2/reporter/reporter.go
Sodman: As far as I'm aware these should always be multierrs or nil... but if either (or both) of them is a non-nil non-multi err, what would be the expected behavior here? Concat the error strings?

Sodman: I guess if one is multi and the other is not, we can add the non-multi to the multi... and if both are just regular errors then we concat?

kdorosh: if they both aren't multis, we could just make a multi

pkg/api/v2/reporter/reporter.go
```diff
@@ -30,9 +30,71 @@ func (e ResourceReports) Accept(res ...resources.InputResource) ResourceReports
...
+			errs1, isFirstMulti := firstReport.Errors.(*multierror.Error)
+			errs2, isSecondMulti := secondReport.Errors.(*multierror.Error)
+
+			// If the errors are not mutliErrs, wrap them in multiErrs:
+			if !isFirstMulti {
+				errs1 = &multierror.Error{Errors: []error{firstReport.Errors}}
+			}
+			if !isSecondMulti {
+				errs2 = &multierror.Error{Errors: []error{secondReport.Errors}}
+			}
```
kdorosh, line 79: we shouldn't need to do this, `multierror.Append` handles this for us already

kdorosh (approved):

PR #399, opened 2020-12-17
New gomock.Matcher

ashleywang1:


test/matchers/proto.go
```diff
@@ -38,6 +67,32 @@ func (p *protoMatcherImpl) NegatedFailureMessage(actual interface{}) (message st
...
+	return len(diff) == 0, nil
+}
+
+func (p *publicFieldMatcher) FailureMessage(actual interface{}) (message string) {
+	return strings.Join(p.diff, " ")
+}
+
+func (p *publicFieldMatcher) NegatedFailureMessage(actual interface{}) (message string) {
+	return "not equal"
+}
```
EItanya, line 94: Can we get better messages for these?

test/matchers/proto.go
```diff
@@ -1,13 +1,42 @@
...
+	"github.com/go-test/deep"
+	"github.com/golang/mock/gomock"
 	"github.com/golang/protobuf/proto"
 	"github.com/onsi/gomega/format"
 	"github.com/onsi/gomega/types"
 )
 
+// Use this in a gomock EXPECT call e.g.
+// `client.EXPECT().Update(ctx, GomockMatchProto(expected)).Return(nil)`
+func GomockMatchProto(actual interface{}) gomock.Matcher {
```
EItanya, line 17: Should we rename this to `GomockMatchPublicFields` since it doesn't only match protos anymore?

EItanya (approved):

PR #403, opened 2021-02-01
add gpl license check to generated-code

saiskee:


ci/oss_compliance/osa_provided.md
```diff
@@ -0,0 +1,44 @@
+Name|Version|License
```
mlholland, line 1: is this the same list as what was used in the equivalent gloo PR, or is it specific to what each repo uses?

saiskee: It is generated per repo

mlholland (approved):

PR #404, opened 2021-02-03
Resource Factories Do Not Register CRDs

sam-heilbron:
# Description

Resource factories will no longer register CRDs

# Context

The code to register CRDs wasn't being used in the first place. But even still there are reasons to get rid of this code:

There are 2 ways that CRDs get registered: through Helm and through our resource factories. The desire behind having both implementations available was ease of use for the user. They could apply yaml and we would register the CRD if it wasn't already there. However, for CRD registration to work via resource factories in production, we need to give the client (in this case Gloo) RBAC permissions to create CRDs. At the moment Gloo doesn't have these permissions (so the code would fail anyway), but more importantly we want Gloo to have minimum permissions.

It turns out that a lot of the kubernetes functionality in solo-kit is most effectively tested when we have the ability to dynamically create/register CRDs. Therefore, I moved the functionality to a test helper and updated any tests that relied on it.

# Testing
- [x] All tests pass
- [x] Open a Gloo PR based off these changes, and ensure all CI tests pass: https://github.com/solo-io/gloo/pull/4280


pkg/api/v1/clients/kube/resource_client_test.go
```diff
@@ -57,16 +57,17 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	kubeClient, err := kubernetes.NewForConfig(cfg)
 	Expect(err).NotTo(HaveOccurred())
 
-	lock, err = clusterlock.NewTestClusterLocker(kubeClient, clusterlock.Options{
+	lock, err = clusterlock.NewKubeClusterLocker(kubeClient, clusterlock.Options{
```
sam-heilbron, line 60: Just moved away from the deprecated API. This is the method the deprecated API was calling

pkg/api/v1/clients/kube/resource_client_test.go
```diff
@@ -57,16 +57,17 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	})
+	Expect(err).NotTo(HaveOccurred())
 	Expect(lock.AcquireLock()).NotTo(HaveOccurred())
 
 	// Create the CRD in the cluster
 	apiExts, err := apiext.NewForConfig(cfg)
 	Expect(err).NotTo(HaveOccurred())
 
-	err = v1.MockResourceCrd.Register(ctx, apiExts)
+	err = helpers.AddAndRegisterCrd(ctx, v1.MockResourceCrd, apiExts)
```
sam-heilbron, line 70: Previously, Crds were automatically added to the registry. Since we removed this (since the registry is just a testing utility), we must add and register the CRD

kdorosh (approved):
looks great! nice comments in code on PR too

PR #405, opened 2021-02-23
Support XDS v3 Client

sam-heilbron:
# Work In Progress
- [ ] Gloo PR (https://github.com/solo-io/gloo/pull/4340) which references this work passing CI

# Description
Migrate xds clients to use v3 envoy (from v2)

# Context
This had issues because we are defining some protos in solo-kit that are already defined in gloo. 

To be specific, the discovery.proto requires some messages defined in `envoy/config/core/v3/base.proto`.
However, Gloo already has this file defined with a supereset of messages.

To avoid moving a bunch of proto messages from the Gloo defined base.proto into the solo-kit defined base.proto, I opted to change the package name of solo-kit defined v3 envoy protos. Instead of using the traditional `solo.io` prefix (which we use across solo-kit and Gloo), I used a `solo.io.kit` prefix. I introduced a README outlining the reasons for this change and the benefits so that it's clear to future developers and can be maintained.

This wasn't an issue in v2 because the v2 discovery client didn't import the v2 base proto, so it was never defined in solo-kit and Gloo at the same time.


pkg/code-generator/codegen/generator.go
```diff
@@ -97,7 +97,7 @@ func GenerateFiles(project *model.Project, skipOutOfPackageFiles, skipGeneratedT
 func generateFilesForXdsResource(resource *model.XDSResource) (code_generator.Files, error) {
 	var v code_generator.Files
 	for suffix, tmpl := range map[string]*template.Template{
-		"_xds.sk.sk.go": templates.XdsTemplate,
+		"_xds.sk.go": templates.XdsTemplate,
```
sam-heilbron, line 100: This was just removing a duplicate `.sk` in the file name

api/external/envoy/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+### v3 Protos
+Some v3 Envoy protos need to be defined in solo-kit (for example, discovery proto messages). We will face package name conflicts if we define solo replicas in multiple projects and attempt to use them at the same time. For example, Gloo uses some solo replicas of the `Envoy.config.core.v3` package. To avoid the package name conflict, we are naming this package `solo.io.kit.[Envoy_PACKAGE]`. (Notice the `kit` in the prefix).
+
+### v3 Discovery Protos
+Discovery protos (ie https://github.com/solo-io/solo-kit/blob/e87adeba6c4dcea75ab88c884d5bef2980c6f5f9/api/external/Envoy/api/v2/discovery.proto) are treated slightly differently. They are purely representational. During code-gen, we use a package replace to swap out our solo defined discovery protos, with the actual go-control-plane defined proto (https://github.com/solo-io/solo-kit/blob/e87adeba6c4dcea75ab88c884d5bef2980c6f5f9/pkg/code-generator/collector/collector.go#L252). Therefore, for discovery protos, we `match the Envoy package exactly`.
```
mlholland, line 11: this link 404s for me, can you change it to standard github file links, rather than blob links?

sam-heilbron: ha I did a find/replace and it capitalized the `e` in envoy. I prefer the blobs (it's the most recent commit in the repo) because if this file changes in the future, the link will always point to the proper place. I've experienced github links pointing to master can point to outdated information when that file/contents change. Are there benefits of the github master links that I'm not considering?

sam-heilbron:
This work was done to support serving rate-limit and ext-auth config using xds v3. Since we will control the transport, we do not need to keep the version up to date with the latest envoy. Instead we will be moving v2 discovery config into solo-kit (from the go-control-plane). THat will be done in a different PR but it means that this PR is no longer necessary.

PR #406, opened 2021-02-24
Support XDS v3 Client (move some protos from gloo)

sam-heilbron:
This had issues becauase we are defining some protos in solo-kit that are already defined in gloo. 

To be specific, the discovery.proto requires some messages defined in envoy/config/core/v3/base.proto.
However, gloo already has this file defined with a supereset of messages.

So either we break up the file and define a base_solo_kit.proto in solo-kit and a base.proto in gloo, or we move the gloo definitions to solo-kit and update the references.

This PR is WIP progress as I want to explore option B so that we follow the envoy proto definition strategy.

This wasn't an issue in v2 because the v2 discovery client didn't import the v2 base proto, so it was never defined in solo-kit.


sam-heilbron:
Closing in favor of: https://github.com/solo-io/solo-kit/pull/405

PR #410, opened 2021-03-09
Set snapshot fix

npolshakova:
Currently, there is no way to determine if a resource in the snapshot does not exist or should be empty. This means SetSnapshot will send a response to envoy with nil listeners/routes if only the endpoints/clusters are set. 

The gloo envoy snapshot will set the resource to an empty list if the intended behavior is for the resources to be cleared. Otherwise when updating only endpoints it will either use the previous snapshot value or not set the listener/route.

Part of fix for: https://github.com/solo-io/gloo/issues/4345

### DIFF
--- a/pkg/api/v1/control-plane/cache/cache_test.go
+++ b/pkg/api/v1/control-plane/cache/cache_test.go
@@ -2,7 +2,6 @@ package cache_test
 
 import (
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
-	rsrc "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/solo-io/solo-kit/pkg/api/v1/control-plane/cache"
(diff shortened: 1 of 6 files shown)


pkg/api/v1/control-plane/cache/simple.go
```diff
@@ -201,7 +206,7 @@ func (cache *snapshotCache) CreateWatch(request Request) (chan Response, func())
 
 	info, ok := cache.status[nodeID]
 	if !ok {
-		info = newStatusInfo(request.Node)
+		info = NewStatusInfo(request.Node)
```
npolshakova, line 209: Make available for the cache test to get status info

pkg/api/v1/control-plane/cache/simple.go
```diff
@@ -139,10 +139,15 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
 				if cache.log != nil {
 					cache.log.Infof("respond open watch %d%v with new version %q", id, watch.Request.ResourceNames, version)
 				}
-				cache.respond(watch.Request, watch.Response, snapshot.GetResources(watch.Request.TypeUrl).Items, version)
 
-				// discard the watch
-				delete(info.watches, id)
+				resources := snapshot.GetResources(watch.Request.TypeUrl).Items
+				if resources != nil {
+					// snapshot has been initialized and exists
```
yuval-k, line 145: let's expand the comment to include the different between nil and zero, as this is main motivation for this change

pkg/api/v1/control-plane/cache/cache_test.go
```diff
@@ -1,9 +1,12 @@
-package cache
+package cache_test
 
 import (
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	rsrc "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
```
EItanya, line 5: Any reason you are using the `go-control-plane` resource package? I think we have these constants defined as well

npolshakova: No I can change it to use the solo-kit resource package constants instead!

yuval-k (approved):

PR #413, opened 2021-03-17
Filter reports by resource kind

jahvon:
Added in this filter so that we could get resources/reports without knowing which resources are in the report. Related to https://github.com/solo-io/gloo/pull/4439

### DIFF
--- a/pkg/api/v2/reporter/reporter.go
+++ b/pkg/api/v2/reporter/reporter.go
@@ -137,6 +137,16 @@ func (e ResourceReports) Find(kind string, ref *core.ResourceRef) (resources.Inp
 	return nil, Report{}
 }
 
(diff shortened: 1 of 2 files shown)


jahvon:
\kick

pkg/api/v2/reporter/reporter.go
```diff
@@ -137,6 +137,16 @@ func (e ResourceReports) Find(kind string, ref *core.ResourceRef) (resources.Inp
 	return nil, Report{}
 }
 
+func (e ResourceReports) FilterByKind(kind string) map[resources.InputResource]Report {
```
npolshakova, line 140: Isn't map[resources.InputResource]Report the same type as ResourceReports?

jahvon: Ah good catch!

npolshakova (approved):
Looks good!

PR #415, opened 2021-04-12
Introduce schemagen

sam-heilbron:
schemagen is a new tool used to generate validation schema (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema) for custom resource definitions. 

A few notes:
- Currently we are using v1beta1 CRDs but that will change in a new version (this version is deprecated and being removed from future versions of kubernetes). I wanted to avoid managing multiple releases of solo-kit, so I made the validation schema interface in a way that we can support multiple versions (where each version has it's specific config). 
- CRDs are managed by Gloo, not solo-kit, so the API needed to support some sort of callback for what to do with the validation schema.


pkg/code-generator/schemagen/generate.go
```diff
@@ -0,0 +1,74 @@
+package schemagen
+
+import (
+	"github.com/hashicorp/go-multierror"
+	"github.com/solo-io/solo-kit/pkg/code-generator/model"
+	"github.com/solo-io/solo-kit/pkg/code-generator/schemagen/v1beta1"
+)
+
+type ValidationSchemaOptions struct {
+	SchemaOptionsByName map[string]v1beta1.SchemaOptions
```
kdorosh, line 10: name of what? resource GVK?

pkg/code-generator/schemagen/v1beta1/test_utils/generated/manifest-list.gen.yaml
```diff
@@ -0,0 +1,57 @@
...
+  versions:
+  - name: v1
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: null
+  storedVersions: null
```
kdorosh, line 57: this is a dup of the one above?

sam-heilbron:
Closing in favor of: https://github.com/solo-io/solo-kit/pull/421

PR #417, opened 2021-04-20
Proto compiler optimizations and metrics

sam-heilbron:
# Description

- When compiling protos, use memoization to avoid parsing the same file multiple times
- Add some basic metrics around duration and frequency of execution, to better understand where bottlenecks are.

# Context

Some recent work in solo-kit (https://github.com/solo-io/solo-kit/pull/415) illustrated that proto compilation (which requires walking a tree of files, accumulating all the imports) takes a long time. This is due to the fact that for each imported file, we parse that file, read the imports and continue. This means, that if we fileA and fileB both import fileC, we will parse fileC twice. The amount of duplicated work grows as the dependency graph gets more complex. 

I added some simple metrics to understand the compilation bottleneck, and ran codegen for gloo. Below are the measurements for our existing implementation of proto compilation compared with the measurements from this change.

### Before (no caching)

```
{
    "duration": {
        "@code-generator/proto-compiler": "23.12s"
    },
    "frequency": {
        "@code-generator/collector-opened-files": 16116
    }
}
```

### After (with caching)

```
{
    "duration": {
        "@code-generator/proto-compiler": "8.41s"
    },
    "frequency": {
        "@code-generator/collector-opened-files": 135
    }
}
```

# Testing
- [x] Successfully executed code-gen in gloo, using this version

### DIFF
--- a/changelog/v0.18.4/proto-compiler-optimization.yaml
+++ b/changelog/v0.18.4/proto-compiler-optimization.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/4559
+    resolvesIssue: false
+    description: Adds memoization to proto compilation to reduce execution time.
--- a/pkg/code-generator/collector/collector.go
+++ b/pkg/code-generator/collector/collector.go
@@ -1,0 +1,154 @@
+package collector
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/solo-io/solo-kit/pkg/code-generator/metrics"
+
+	"github.com/rotisserie/eris"
+	"github.com/solo-io/go-utils/log"
+	"github.com/solo-io/go-utils/stringutils"
+)
+
+type Collector interface {
+	CollectImportsForFile(root, protoFile string) ([]string, error)
+}
+
+func NewCollector(customImports, commonImports []string) *collector {
+	return &collector{
+		customImports:    customImports,
+		commonImports:    commonImports,
+		importsExtractor: NewSynchronizedImportsExtractor(),
+	}
+}
+
+type collector struct {
+	customImports []string
+	commonImports []string
+
+	// The collector traverses a tree of files, opening and parsing each one.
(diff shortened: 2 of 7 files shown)


changelog/v0.18.4/proto-compiler-optimization.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NON_USER_FACING
```
kdorosh, line 2: breaking change?

pkg/code-generator/collector/collector.go
```diff
@@ -3,153 +3,51 @@ package collector
...
 	"github.com/solo-io/go-utils/stringutils"
-	"github.com/solo-io/solo-kit/pkg/code-generator/model"
-	"github.com/solo-io/solo-kit/pkg/code-generator/parser"
-	"github.com/solo-io/solo-kit/pkg/errors"
-	"golang.org/x/sync/errgroup"
 )
 
 type Collector interface {
-	CollectDescriptorsFromRoot(root string, skipDirs []string) ([]*model.DescriptorWithPath, error)
+	CollectImportsForFile(root, protoFile string) ([]string, error)
```
kdorosh, line 19: this makes for a breaking change

pkg/code-generator/collector/collector.go
```diff
@@ -181,15 +88,14 @@ func (c *collector) importsForProtoFile(absoluteRoot, protoFile string, customIm
...
 		if err != nil {
-			return nil, errors.Wrapf(err, "getting imports for dependency")
+			return nil, eris.Wrapf(err, "getting imports for dependency")
 		}
 		importsForProto = append(importsForProto, strings.TrimSuffix(importPath, "/"))
 		importsForProto = append(importsForProto, dependencyImports...)
 	}
-
-	return importsForProto, nil
+	return stringutils.Unique(importsForProto), nil
```
kdorosh, line 98: this feels like a behavior change. how did we handle duplicates before?

sam-heilbron: We used to do this further out, in the compiler: https://github.com/solo-io/solo-kit/pull/417/files#diff-c1b7cbb4edcfe824cf0146bf0190676e1b9b2d7489ddcb828f082c45b1b5928aL103

kdorosh (approved):

PR #421, opened 2021-05-14
schemagen: generate kube validation schemas with structural schema constraints

sam-heilbron:
# Description
schemagen is a new tool used to generate validation schema (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema) for custom resource definitions.

# Schemagen 

## How are schemas generated?
At a high level, schemagen follows these steps:
- User defines a path to a directory where CRDs are stored in files
- User defines for each project, whether or not to generate validation schemas
- For each project, if schemas are to be generated:
    1. The CRDs are parsed from the source directory and stored in a map, keyed by Group-Version-Kind.
    2. A JsonSchemaGenerator creates JsonSchemas for each resource in the project, and stores them in a map, keyed by Group-Version-Kind of the resource. For details on how the JsonSchemaGenerators work, see the below sections.
    3. Schemas are matched to CRDs, using the Group-Version-Kind to match them, and then written back to the source directory with the following naming pattern: `[GROUP]_[KIND].yaml`
 
## JsonSchemaGenerator: Cuelang implementation
`Use cuelang as an intermediate language for transpiling protobuf schemas to openapi v3 with k8s structural schema constraints.`

This is our preferred implementation. It is not used in production yet, due to some performance issues with the Gloo Edge API. However, our goal is to eventually migrate our code to rely on this implemenation. I added in this code in this iteration so that we could compare the generated schemas between the implementations. 

https://github.com/cuelang/cue/discussions/944 tracks the issue we face when using cuelang with the Gloo Edge API. Specifically, as more oneof's are added to the API, the performance degrades dramatically, to the point where it takes longer than 15 minutes to generate schemas. The authors are aware of this issue and are actively working on improving it. Until it is resolved, we cannot move forward with this implementation.

This is the pattern that skv2 relies on (https://github.com/solo-io/skv2/blob/master/codegen/render/manifests_renderer.go) and is loosely based on an istio tool `cue-gen` (https://github.com/istio/tools/tree/master/cmd/cue-gen). These API's don't have the same number of oneof's as Gloo Edge and therefore don't face the same performance degredation.

## JsonSchemaGenerator: Protobuf compiler plugin implementation
`protoc-gen-openapi is a plugin for the Google protocol buffer compiler to generate openAPI V3 spec for any given input protobuf. It runs as a protoc-gen- binary that the protobuf compiler infers from the openapi_out flag.`

This is our current implementation that is used in production, but we hope only temporarily. As described above, the latest version of cue doesn't work with the Gloo Edge API for performance reasons. 

This implemenation relies on the protobuf compiler being run with an `openapi_out` flag. For each resource proto, we run protoc with the corresponding flags, write the schemas to a temp file, parse that file and convert it into a JsonSchema.

We rely on an implemenation based on an istio tool: https://github.com/istio/tools/tree/593a41c76c5c84a4cd51a4ab0c345630c5ed30ba/openapi/protoc-gen-openapi. It has since been removed from the repo, but I cloned it, and made necessary modifications to support specific Gloo protos. 

# Testing Checklist
- [x] Convert a simple resource (`testing.solo.io.SimpleMockResource`) into JsonSchema using the protoc implementation and cuelang implemenation, and validate they produce similar structures. 
- [x] Run codegen in gloo, and ensure that it completes. There will be some additional work to validate that the Gloo API is properly converted into open api schemas but for this PR, I want to ensure at a minimum that schemagen completes, even if it doesn't produce a production-ready validation schema. (Gloo Edge PR: https://github.com/solo-io/gloo/pull/4757)


generate.go
```diff
@@ -25,6 +25,11 @@ func main() {
 				sk_anyvendor.EnvoyValidateProtoMatcher.Package: sk_anyvendor.EnvoyValidateProtoMatcher.Patterns,
 			},
 		},
+		ValidationSchemaOptions: &schemagen.ValidationSchemaOptions{
+			// Path to where test CRDs are stored
+			CrdDirectory:   "test/mocks/crds",
+			JsonSchemaTool: "protoc",
```
sam-heilbron, line 31: It's probably unnecessary to expose this to the user. I had done this to make testing easier while I was switching back and forth between the implementations. Open to thoughts

Makefile
```diff
@@ -38,6 +38,7 @@ update-deps: mod-download
 	mkdir -p $(DEPSGOBIN)
 	chmod +x $(shell go list -f '{{ .Dir }}' -m k8s.io/code-generator)/generate-groups.sh
 	GOBIN=$(DEPSGOBIN) go install github.com/solo-io/protoc-gen-ext
+	GOBIN=$(DEPSGOBIN) go install github.com/sam-heilbron/protoc-gen-openapi
```
sam-heilbron, line 41: For testing I just created this repo, based on https://github.com/istio/tools/tree/593a41c76c5c84a4cd51a4ab0c345630c5ed30ba/openapi/protoc-gen-openapi. I'm happy to create a solo owned repo to house this code. Our goal is that it is short lived and deleted when we support cue.

pkg/code-generator/schemagen/generator.go
```diff
@@ -0,0 +1,125 @@
...
+			Version: crd.Spec.Version,
+			Kind:    crd.Spec.Names.Kind,
+		}
+
+		specJsonSchema, ok := jsonSchemasByGVK[crdGVK]
+		if !ok {
+			continue
+		}
+
+		removeProtoMetadataValidation(specJsonSchema)
```
sam-heilbron, line 82: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema ... `if metadata is specified, then only restrictions on metadata.name and metadata.generateName are allowed.`

The kube api server is responsible for managing this field, so users are not allowed to define schemas on it. We remove validation altogether.

generate.go
kdorosh: I think it's fine to leave exposed

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/solo-io/solo-kit
 go 1.16
 
 require (
+	cuelang.org/go v0.3.2
```
kdorosh, line 6: why this version, as opposed to 0.4 or 0.2.2?

pkg/code-generator/schemagen/crd.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/solo-io/solo-kit/pkg/code-generator/writer"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/utils/pointer"
+
+	"github.com/ghodss/yaml"
+	kubeyaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+const (
+	apiVersion = "apiextensions.k8s.io/v1beta1"
```
kdorosh, line 24: rename to v1beta1? might even be able to grab this constant from a k8s dependency instead

pkg/code-generator/schemagen/crd.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return c.fileWriter.WriteFile(code_generator.File{
+		Filename: c.getFilenameForCRD(crd),
+		Content:  string(crdBytes),
+	})
+}
+
+func (c *CrdWriter) getFilenameForCRD(crd v1beta1.CustomResourceDefinition) string {
+	return fmt.Sprintf("%s_%s.yaml", crd.Spec.Group, crd.Spec.Names.Kind)
```
kdorosh, line 66: worth including the version to get the entire GVK in the filename?

pkg/code-generator/schemagen/crd.go
```diff
@@ -0,0 +1,123 @@
...
+
+func getCRDFromFile(pathToFile string) (v1beta1.CustomResourceDefinition, error) {
+	crd := v1beta1.CustomResourceDefinition{}
+
+	r, err := os.Open(pathToFile)
+	if err != nil {
+		return crd, err
+	}
+	defer func() {
+		_ = r.Close()
```
kdorosh, line 105: log error if fail?

pkg/code-generator/schemagen/generator.go
kdorosh: add this comment to code?

pkg/code-generator/schemagen/protoc.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		// remove "properties" value
+		delete(values, "properties")
+		// remove "required" value
+		delete(values, "required")
+		// x-kubernetes-preserve-unknown-fields allows for unknown fields from a particular node
+		// see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema
+		values["x-kubernetes-preserve-unknown-fields"] = true
+	}
+}
```
kdorosh, line 211: move shared logic to a shared package?

test/mocks/crds/testing.solo.io_MockResource.yaml
```diff
@@ -0,0 +1,77 @@
+# Code generated by solo-kit. DO NOT EDIT.
```
kdorosh, line 1: 🎉

pkg/code-generator/schemagen/crd.go
sam-heilbron: Yeah I had opted out of including the version, since many versions can be defined on the CRD itself. However, doesn't hurt to add it, can always rename later.

pkg/code-generator/schemagen/protoc.go
sam-heilbron: Happy to! They're slightly different implementations (one is looking for type_url, the other for @type). Though that can easily be handled

go.mod
sam-heilbron: I can try with 0.4. 0.2.2 unfortunately doesn't include message level options which we need in Gloo Edge (added here: https://github.com/cuelang/cue/commit/3af26837fad469c6f5ae98978724691676685d9b)

Makefile
npolshakova: Is this the repo removed from istio? https://github.com/istio/tools/tree/593a41c76c5c84a4cd51a4ab0c345630c5ed30ba/openapi/protoc-gen-openapi  What did they replace it with?

sam-heilbron: They use `cue` https://github.com/istio/tools/tree/master/cmd/cue-gen

pkg/code-generator/schemagen/crd.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	return c.fileWriter.WriteFile(code_generator.File{
+		Filename: c.getFilenameForCRD(crd),
+		Content:  string(crdBytes),
+	})
+}
+
+func (c *CrdWriter) getFilenameForCRD(crd apiextv1beta1.CustomResourceDefinition) string {
```
npolshakova, line 65: This seems more like a helper function than related to the CrdWriter?

pkg/code-generator/schemagen/cue.go
```diff
@@ -0,0 +1,188 @@
...
+
+		return c.convertOpenApiSchemasToJsonSchemas(project, openApiSchemas)
+	}
+
+	return nil, nil
+}
+
+func (c *cueGenerator) getProtobufExtractorForProject(project *model.Project) (*protobuf.Extractor, error) {
+	// Collect all protobuf definitions including transitive dependencies.
+	var imports []string
```
npolshakova, line 99: Should this function be inside the collector package instead of part of the cueGenerator? It looks like CollectImportsForFile does a similar thing for a specific file and this is just collect the protobuf definitions for a project?

pkg/code-generator/schemagen/cue.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	return protobufExtractor, nil
+}
+
+func (c *cueGenerator) convertOpenApiSchemasToJsonSchemas(project *model.Project, schemas OpenApiSchemas) (map[schema.GroupVersionKind]*apiextv1beta1.JSONSchemaProps, error) {
+	jsonSchemasByGVK := make(map[schema.GroupVersionKind]*apiextv1beta1.JSONSchemaProps)
+
+	for schemaKey, schemaValue := range schemas {
+		schemaGVK := c.getGVKForSchemaKey(project, schemaKey)
```
npolshakova, line 130: Can getGVKForSchemaKey ever error here? It looks like it just parses the schemaKey, but it also gets the group and version?

pkg/code-generator/schemagen/crd.go
```diff
@@ -0,0 +1,126 @@
...
+
+	err := filepath.Walk(crdDirectory, func(crdFile string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		if !strings.HasSuffix(crdFile, ".yaml") {
```
npolshakova, line 80: Is there a consistent ".yaml" suffix? I've sometimes seen .yml but is that used interchangeably with .yaml?

pkg/code-generator/schemagen/cue.go
```diff
@@ -0,0 +1,188 @@
...
+		if err != nil {
+			return nil, eris.Errorf("Cue openapi generation failed for %s: %+v", project.ProtoPackage, err)
+		}
+
+		// Iterate openapi objects to construct mapping from proto message name to openapi schema
+		for _, kv := range schemas.Pairs() {
+			openApiSchemas[kv.Key] = kv.Value.(*openapi.OrderedMap)
+		}
+
+		return c.convertOpenApiSchemasToJsonSchemas(project, openApiSchemas)
```
npolshakova, line 91: Is there only one builtInstance in built?

sam-heilbron: No I believe each builtinstance corresponds to a protobuf message.

pkg/code-generator/schemagen/crd.go
sam-heilbron: Good call! I can add support for both.

pkg/code-generator/schemagen/cue.go
sam-heilbron: It shouldn't. If it does error, it's because the project has been misconfigured

pkg/code-generator/schemagen/cue.go
sam-heilbron: I tried to isolate our dependency on cue into a single file to keep it distinct from the rest of our codebase. It's unfortunate that it does something quiet similar to our collector package, but I think we need it because it generates the cue types that we then parse

pkg/code-generator/schemagen/crd.go
sam-heilbron: Fair enough!

pkg/code-generator/schemagen/cue.go
npolshakova: Is it returning early here then? It looks like it returns on the first builtInstance?

sam-heilbron: We skip all the irrelevant protos (https://github.com/solo-io/solo-kit/pull/421/files/4831a802ecb7c520a85777b81bef1f4fac22f189#diff-06c6f9e953ed44cf5c7fea458dfd7ebab8492859838751cb5a098d05dd010465R70) so when we come across one, it means that is the top level proto message (ie Gateway, VirtualService, Proxy ...etc)

pkg/code-generator/schemagen/crd.go
kdorosh: seconding this

kdorosh (commented):
lgtm, leaving final review for @npolshakova

pkg/code-generator/schemagen/crd.go
```diff
@@ -0,0 +1,126 @@
...
+	}()
+
+	f := bufio.NewReader(r)
+	decoder := kubeyaml.NewYAMLReader(f)
+
+	doc, err := decoder.Read()
+	if err != nil {
+		return crd, err
+	}
+	chunk := bytes.TrimSpace(doc)
```
saiskee, line 118: Just curious, is this TrimSpace needed? Will `yaml.Unmarshal` fail if there is trailing spaces?

sam-heilbron: To be honest, i'm not positive. I had grabbed this from the istio implementation here: https://github.com/istio/tools/blob/477454adf7995dd3070129998495cdc8aaec5aff/cmd/cue-gen/test/crdgen_test.go#L67

npolshakova (approved):
Looks good!

PR #422, opened 2021-05-14
Codegen writer

sam-heilbron:
# Description

- Add a code-generator.Writer type to consolidate writing solo-kit generated files
- Convert some info level logging to debug to reduce the verbosity of the code-gen logs
- Split out a type that manages proto execution and introduce an instance that will be used with openapi_out. We will use this new type when generating validation schemas. 
- Rename out generated xds resource files to deduplicate a `.sk` suffix. 

# Context

As part of the work to introduce schema generation, I made these changes. However, I wanted to limit the scope of the schemagen PR, so I broke these changes off into their own PR.

# Testing
- [x] Run codegen in Gloo. The only change should be that the generated xds resources no longer have `.sk.sk.go` suffix and instead just a `.sk.go` suffix


pkg/code-generator/collector/compiler.go
```diff
@@ -153,96 +138,3 @@ func readDescriptors(fromFile string) (*descriptor.FileDescriptorSet, error) {
 	}
 	return &desc, nil
 }
-
-func (c *protoCompiler) findImportRelativeToRoot(absoluteRoot, importedProtoFile string, customImports, existingImports []string) (string, error) {
```
sam-heilbron, line 157: This was a duplicate function that was not being used.

pkg/code-generator/collector/compiler.go
```diff
@@ -153,96 +138,3 @@ func readDescriptors(fromFile string) (*descriptor.FileDescriptorSet, error) {
...
-	return possibleImportPaths[0], nil
-
-}
-
-var defaultGoArgs = []string{
-	"plugins=grpc",
-	"Mgithub.com/solo-io/solo-kit/api/external/envoy/api/v2/discovery.proto=github.com/envoyproxy/go-control-plane/envoy/api/v2",
-}
-
-func (c *protoCompiler) writeDescriptors(protoFile, toFile string, imports []string, compileProtos bool) error {
```
sam-heilbron, line 217: I now have 2 implementations of protoc, one without openapi and one with it. Therefore, I split the implementations into their own file.

pkg/code-generator/codegen/generator.go
```diff
@@ -75,12 +71,7 @@ func GenerateFiles(project *model.Project, skipOutOfPackageFiles, skipGeneratedT
 		}
 		files = append(files, fs...)
 	}
-	for i, v := range files {
```
sam-heilbron, line 78: This PR consolidates writing content and headers into a single Writer type. Therefore, this code can be removed

pkg/code-generator/docgen/generator.go
```diff
@@ -32,26 +34,26 @@ var ignoredFiles = []string{
...
+			DocsDir:     filepath.Join(absoluteRoot, project.ProjectConfig.DocsDir),
 		}
+
+		return GenerateAndWriteFiles(docGenerator)
 	}
 	return nil
 }
 
+// DEPRECATED
+// prefer GenerateAndWriteFiles
```
jahvon, line 56: Is this still being used somewhere?

sam-heilbron: It's not, but because it's a public function, I didn't want to delete it.

jahvon (approved):

PR #425, opened 2021-05-25
upgrade deps

jenshu:


go.mod
```diff
@@ -46,19 +46,18 @@ require (
 	github.com/sam-heilbron/protoc-gen-openapi v0.0.0-20210521134209-c323189e7a9a
 	github.com/solo-io/anyvendor v0.0.1
 	github.com/solo-io/go-list-licenses v0.0.4
-	github.com/solo-io/go-utils v0.20.2
-	github.com/solo-io/k8s-utils v0.0.1
+	github.com/solo-io/go-utils v0.21.6-0.20210525213056-a63ad8750c56
+	github.com/solo-io/k8s-utils v0.0.8-0.20210525213538-c36d2ee1d56f
```
jenshu, line 50: TODO: update when these are released

pkg/api/v1/clients/kube/cache/cache_test.go
```diff
@@ -44,7 +44,7 @@ var _ = Describe("kube core cache tests", func() {
 			client = kubernetes.NewForConfigOrDie(cfg)
 		})
 
-		Context("all namesapces", func() {
+		Context("all namespaces", func() {
```
sam-heilbron, line 47: haha when I was looking at this test, I kept getting bothered by this so thanks for fixing!

pkg/api/v1/clients/kube/cache/cache_test.go
```diff
@@ -166,3 +168,14 @@ var _ = Describe("kube core cache tests", func() {
 func init() {
 	rand.Seed(time.Now().UnixNano())
 }
+
+// remove the auto-generated kube-root-ca.crt ConfigMap, if it exists
+func cleanConfigMaps(configMaps []*v1.ConfigMap) []*v1.ConfigMap {
```
sam-heilbron, line 173: Do we know why this suddenly started getting generated?

jenshu: unfortunately not. from looking at the code, it seems like it should have always been generated, but not sure why this only started popping up yesterday

saiskee (approved):

PR #426, opened 2021-06-02
Go package compliance

jehawley:
New versions of `github.com/golang/protobuf` require that the `go_package` in proto files has at least one '/'. 

This PR updates the offending protos from what is, as far as I can tell, their current versions:
https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto
https://github.com/prometheus/client_model/blob/master/metrics.proto

In theory, these path updates are breaking changes (hence the 0.21.0 version). In practice, it does not seem that any go code currently imports them (that is, searching the solo-io org on github did not produce any hits for the go package name in go code).
BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/427

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ root level `generate.go` file.
 
 ## upgrading to v0.12.0 (solo-kit with go.mod)
 
-As of go 1.11, go began introducing support for go modules, it's dependency management system.
+As of go 1.11, go began introducing support for go modules, its dependency management system.
 As of solo-kit 0.12.0 we will officially support running solo-kit with go.mod outside of the GOPATH.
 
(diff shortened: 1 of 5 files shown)


marcogschmidt (commented):
Looks good to me. Did you check whether there are other proto files in this repo that have the same problem?

jehawley:
> Looks good to me. Did you check whether there are other proto files in this repo that have the same problem?

Yes - there are no others.  `ag 'go_package[^/]+;$' ./` returns only these 2 files (and none after they are fixed)

marcogschmidt (approved):

PR #423, opened 2021-05-17
Clean up Envoy V2 references

jahvon:
This PR pulls in a few Envoy V2 api protos into Solo Kit and completely removes all go-control-plane v2 API references. Protos that were pulled in from Envoy were slightly modified so that they work with our proto generation (including adding in the go package definition, importing extproto, and updating the package names). 

Since we no longer have to support the v2 API for listeners/routes/etc., those resources and related code were deleted from solo-kit. The Envoy ADS server was also repurposed as a "Solo Discovery Service" since we only need that for our internal communication from ext-auth and rate-imit.

We are doing this since we use the V2 API in Gloo but it is likely that they will be removed from `go-control-plane` (where we are currently pulling them in from)

**Related PRs**
Gloo change to pull in these APIs instead of go-control-plane https://github.com/solo-io/gloo/pull/4777
SP changes to pull in the updated Gloo & SK https://github.com/solo-io/solo-projects/pull/2347

PR #428, opened 2021-06-02
Undo discovery proto package name change

jahvon:
This PR reverts [this](https://github.com/solo-io/solo-kit/pull/423/files#diff-58de2eafe6af4e2435bb78bdc8cbc1fe8956ed5a5b7859a0bd12f2cd36f3d839R3) change (discovery proto package rename)


api/external/envoy/api/v2/discovery.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package solo.io.envoy.api.v2;
```
jahvon, line 3: This rename could result in downtime for ext-auth/rate-limit so it's being reverted

saiskee (approved):

PR #429, opened 2021-06-25
Add support for statuses from multiple reporters

MitchAman:
Add support for having statuses from more than one controller.  This PR adds a new proto message, `NamespacedStatuses`, that acts as a replacement for `Status`.  `NamespacedStatuses` is a mapping from `POD_NAMESPACE` (env var) to `Status`.  This allows multiple gloo instances in different namespaces to record their own statuses without overwriting one another. 

**Backwards Compatibility**
- As done in the mock resources in this PR, existing resources containing a `Status` field will need to move that field into a new `oneof` containing a `NamespacedStatuses` as well.
- `GetStatus()` and `SetStatus(Status)` functions on resources should no longer be used. Instead use `GetNamespacedStatus()` and `UpsertNamespacedStatuses(Status)`, respectively.
- In order for `GetNamespacedStatus()` and `UpsertNamespacedStatuses(Status)` to behave properly, the `POD_NAMESPACE` environment variable **must** be set.

BOT NOTES: 
resolves https://github.com/solo-io/solo-kit/issues/430


api/v1/status.proto
```diff
@@ -8,6 +8,14 @@ option (extproto.equal_all) = true;
 
 import "google/protobuf/struct.proto";
 
+/**
+ * ReporterStatus indicates the Status of a resource according to each reporter
+ */
+message ReporterStatus {
+    // Mapping from reporter (Status.reported_by) to that reporter's Status
+    map<string, Status> statuses = 1;
```
kdorosh, line 16: let's add a controller ID or hash to a status so we know who wrote it in the single controller case. perhaps we can grab the pod revision and use that from env? and also use it as the key to map here

MitchAman: Per our offline discussion I've modified this to use the pod's namespace

pkg/api/v2/reporter/reporter.go
```diff
@@ -277,9 +277,8 @@ func attemptUpdateStatus(ctx context.Context, client ReporterResourceClient, res
 		//    Also, the status is accurate for the resource as it's stored in Gloo's memory in the interim.
 		//    This is explained further here: https://github.com/solo-io/solo-kit/pull/360#discussion_r433397163
 		if inputResourceFromRead, ok := resourceFromRead.(resources.InputResource); ok {
-			status := resourceToWrite.GetStatus()
 			resourceToWrite = inputResourceFromRead
-			resourceToWrite.SetStatus(status)
+			resourceToWrite.AddToReporterStatus(status)
```
sam-heilbron, line 281: Do we want to call SetStatus as well here?

sam-heilbron (commented):
This is looking great!

Main lingering question is just around hw we want to maintain the Status and ReporterStatus. Will only multi-namespace resources write to the ReporterStatus? Or will all resources update both?

Now that we're breaking the single-writer principle, is there any concern of multiple controllers stepping on one another when updating the ReporterStatus? Is it worth it to write a test with concurrent controllers?

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -457,8 +459,28 @@ func (rc *ResourceClient) convertCrdToResource(resourceCrd *v1.Resource) (resour
...
+				typedStatus := core.ReporterStatus{}
+				if err := protoutils.UnmarshalMapToProto(resourceCrd.Status, &typedStatus); err != nil {
+					return err
+				}
+				*status = typedStatus
+				return nil
+			}
+			// First attempt to unmarshal Status
+			if statusErr := resources.UpdateStatus(withStatus, updateStatusFunc); statusErr != nil {
+				// If unmarshalling Status failed, the resource likely has a ReporterStatus instead.
```
kdorosh, line 475: let's switch the order here? going forward, it will be more likely we see a reporter status, so no reason to try and fail to unmarshal it as a status all the time. slightly more efficient on the cpu in more cases

pkg/api/v1/clients/kube/resource_client_test.go
```diff
@@ -119,6 +119,13 @@ var _ = Describe("Test Kube ResourceClient", func() {
 	BeforeEach(func() {
 		ctx = context.Background()
 		client = kube.NewResourceClient(v1.MockResourceCrd, clientset, kube.NewKubeCache(ctx), &v1.MockResource{}, []string{metav1.NamespaceAll}, 0)
+		err := os.Setenv("POD_NAMESPACE", "test-ns")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		err := os.Setenv("POD_NAMESPACE", "")
```
kdorosh, line 127: there is an `os.UnsetEnv()`

MitchAman: Thanks, didn't realize that!

pkg/api/v1/reconcile/reconciler.go
```diff
@@ -109,7 +109,8 @@ func updateDesiredResourceVersionAndStatus(desired, original resources.Resource)
 		meta.ResourceVersion = original.GetMetadata().ResourceVersion
 	})
 	if desiredInput, ok := desired.(resources.InputResource); ok {
-		desiredInput.SetStatus(&core.Status{})
+		// TODO(mitchaman): remove hardcoded gateway
```
kdorosh, line 112: address todo

pkg/api/v1/reporter/reporter_test.go
```diff
@@ -42,12 +53,15 @@ var _ = Describe("Reporter", func() {
 		Expect(err).NotTo(HaveOccurred())
 		r2, err = mockResourceClient.Read(r2.GetMetadata().Namespace, r2.GetMetadata().Name, clients.ReadOpts{})
 		Expect(err).NotTo(HaveOccurred())
-		Expect(r1.(*v1.MockResource).GetStatus()).To(Equal(&core.Status{
-			State:      2,
-			Reason:     "everyone makes mistakes",
-			ReportedBy: "test",
-		}))
-		Expect(r2.(*v1.MockResource).GetStatus()).To(Equal(&core.Status{
+		fmt.Println(r1.(*v1.MockResource).GetReporterStatus().String())
```
kdorosh, line 56: remove println

pkg/code-generator/codegen/templates/kube/types_template.go
```diff
@@ -59,13 +60,15 @@ func (o *{{ .Name }}) MarshalJSON() ([]byte, error) {
...
 	delete(spec, "status")
+	delete(spec, "reporter_status")
 {{- end }}
 	asMap := map[string]interface{}{
 		"metadata":   o.ObjectMeta,
 		"apiVersion": o.TypeMeta.APIVersion,
 		"kind":       o.TypeMeta.Kind,
 {{- if .HasStatus }}
 		"status":     o.Status,
+		"reporter_status": o.ReporterStatus,
```
kdorosh, line 71: goimports formatting here could be a bit nicer

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -87,8 +87,67 @@ func (r *{{ .Name }}) SetMetadata(meta *core.Metadata) {
...
+	}
+}
+
+// GetStatusForReporter returns the status stored in the ReporterStatus.Statuses map for the
+// controller specified by reportedBy, or nil if no status exists for that controller.
+// Note: POD_NAMESPACE environment variable must be set for this function to behave as expected.
+func (r *{{ .Name }}) GetStatusForReporter(reportedBy string) *core.Status {
+	podNamespace := os.Getenv("POD_NAMESPACE")
+	if podNamespace != "" {
+		key := podNamespace + ":" + reportedBy
```
kdorosh, line 121: the old `GetStatus()` didn't require `reportedBy` to be provided by the user. we can probably simplify this to avoid passing in the arg here since `reportedBy` today will always be the same controller for gloo

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -87,8 +87,67 @@ func (r *{{ .Name }}) SetMetadata(meta *core.Metadata) {
...
+		}
+		key := podNamespace + ":" + status.GetReportedBy()
+		r.GetReporterStatus().Statuses[key] = status
+	}
+}
+
+// GetStatusForReporter returns the status stored in the ReporterStatus.Statuses map for the
+// controller specified by reportedBy, or nil if no status exists for that controller.
+// Note: POD_NAMESPACE environment variable must be set for this function to behave as expected.
+func (r *{{ .Name }}) GetStatusForReporter(reportedBy string) *core.Status {
```
kdorosh, line 118: perhaps rename to `GetNamespacedStatus()`?

pkg/api/v1/resources/resource_interface.go
```diff
@@ -46,6 +46,12 @@ type InputResource interface {
 	Resource
 	GetStatus() *core.Status
 	SetStatus(status *core.Status)
+	HasStatus() bool
+	GetReporterStatus() *core.ReporterStatus
+	SetReporterStatus(status *core.ReporterStatus)
+	HasReporterStatus() bool
+	GetStatusForReporter(reportedBy string) *core.Status
+	AddToReporterStatus(status *core.Status)
```
kdorosh, line 54: this functions more like an upsert than an add, lets rename?

pkg/code-generator/codegen/templates/resource_template.go
MitchAman: The `reportedBy` field is needed for the key in the `ReporterStatus.Statuses` map.  Or are you suggesting we change that map key from "POD_NAMESPACE:reportedBy" to just "POD_NAMESPACE"?

kdorosh: I'm just noting that the original function didn't have any params and it still said "reported_by: gloo" or "reported_by: gateway" and that we should be able to lean on that

MitchAman: There is still the `GetReporterStatus()` variation which does not have any params, but the caller would then need to locate the `Status` in the map that's relevant to them.  The idea behind `GetStatusForReporter(reportedBy string)` is to be a helper function so that each caller doesn't need to know the map key format, and can instead get the `Status` they care about directly.

pkg/api/v2/reporter/reporter.go
MitchAman: (resolving outdated)

pkg/api/v1/clients/kube/resource_client.go
MitchAman: Good idea, thanks

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -457,8 +459,28 @@ func (rc *ResourceClient) convertCrdToResource(resourceCrd *v1.Resource) (resour
...
+				typedStatus := core.ReporterStatus{}
+				if err := protoutils.UnmarshalMapToProto(resourceCrd.Status, &typedStatus); err != nil {
+					return err
+				}
+				*status = typedStatus
+				return nil
+			}
+			// First attempt to unmarshal ReporterStatus
+			if reporterStatusErr := resources.UpdateReporterStatus(withStatus, updateReporterStatusFunc); reporterStatusErr != nil {
+				// If unmarshalling ReporterStatus failed, the resource likely has a Status instead.
```
saiskee, line 475: Can we implement a check on the error message to make sure that the error is what we expect here (the status is of type `Status` rather than `ReporterStatus`), and not another k8s error?

saiskee: I see down below you report both errors if unmarshalling to `Status` fails. This isn't necessary then, but a nit. We might want to just catch and return the error earlier on if it's not an `Unmarshal` error.

pkg/api/v1/reporter/reporter_test.go
```diff
@@ -19,6 +21,15 @@ var _ = Describe("Reporter", func() {
 		reporter                               rep.Reporter
 		mockResourceClient, fakeResourceClient clients.ResourceClient
 	)
+
+	BeforeEach(func() {
+		Expect(os.Setenv("POD_NAMESPACE", "gloo-system")).NotTo(HaveOccurred())
```
saiskee, line 26: Why do we set this? Is this normally set by k8s?

MitchAman: Yes, it comes from k8s, we plumb it through in our deployment templates, i.e.: https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/1-gloo-deployment.yaml#L169-L172

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -87,8 +87,66 @@ func (r *{{ .Name }}) SetMetadata(meta *core.Metadata) {
...
+	r.StatusOneof = &{{ .Name }}_ReporterStatus{ReporterStatus: status}
+}
+
+// UpsertReporterStatus inserts the specified status into the ReporterStatus.Statuses map for the
+// current namespace (as specified by POD_NAMESPACE env var).  If the resource does not yet have
+// a ReporterStatus, one will be created.
+// Note: POD_NAMESPACE environment variable must be set for this function to behave as expected.
+func (r *{{ .Name }}) UpsertReporterStatus(status *core.Status) {
+	podNamespace := os.Getenv("POD_NAMESPACE")
+	if podNamespace != "" {
```
saiskee, line 103: Should we be doing something here if `POD_NAMESPACE` env var is not set? Some sort of error reporting? This can silently fail if os.getenv(POD_NAMESPACE) fails to get the pod namespace..

pkg/api/v1/resources/resource_interface.go
```diff
@@ -44,8 +44,16 @@ func ProtoCast(res Resource) (ProtoResource, error) {
...
 	Resource
+	// Deprecated: statuses are now written to a ReporterStatus, use GetNamespacedStatus() or
+	// GetReporterStatus() instead.
 	GetStatus() *core.Status
 	SetStatus(status *core.Status)
+	HasStatus() bool
+	GetReporterStatus() *core.ReporterStatus
+	SetReporterStatus(status *core.ReporterStatus)
+	HasReporterStatus() bool
+	GetNamespacedStatus() *core.Status
```
sam-heilbron, line 55: What's the purpose of expanding the interface for HasReporterStatus and UpsertReporterStatus? I'm a little confused by deprecating methods on the existing interface to prefer new ones with the same signature

test/mocks/crds/testing.solo.io_v1_MockResource.yaml
```diff
@@ -57,15 +57,23 @@ spec:
...
+                    type: object
+                    x-kubernetes-preserve-unknown-fields: true
+                  description: Mapping from namespace to the Status written by the
+                    controller running in that namespace.
+                  type: object
+              type: object
             someDumbField:
               type: string
             status:
-              description: The status field comment
```
sam-heilbron, line 63: We have a status field on the spec, and a status resource on the CRD. What is the difference between the two? They seem to be generated from the same field, why is there not a duplicate reporterStatus as well?

pkg/code-generator/codegen/templates/resource_template.go
MitchAman: Good idea, this wouldn't be a fun issue to track down. Going to do the same for `GetNamespacedStatus`

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -457,8 +459,28 @@ func (rc *ResourceClient) convertCrdToResource(resourceCrd *v1.Resource) (resour
...
+				statusErr := resources.UpdateStatus(withStatus, updateStatusFunc)
```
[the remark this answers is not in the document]
MitchAman: Correct.  Yes, we do want re-write the status as a ReporterStatus. We are actually doing that in a roundabout way.  `resources.UpdateStatus` calls `SetStatus` [1], which has been modified to write a `ReporterStatus` [2]

[1] https://github.com/solo-io/solo-kit/blob/ec7dcf636b846f4cdb1897f9886f589c6ff150a2/pkg/api/v1/resources/resource_interface.go#L465
[2] https://github.com/solo-io/solo-kit/blob/ec7dcf636b846f4cdb1897f9886f589c6ff150a2/pkg/code-generator/codegen/templates/resource_template.go#L90

pkg/api/v1/resources/resource_interface.go
MitchAman: `HasReporterStatus` is a helper for determining if the `status_oneof` contains a `ReporterStatus`, as opposed to a `Status`.  It's necessity is debatable, and usages could probably be replaced by `.GetReporterStatus() != nil`, but I kind of like the Has...() syntax as a subtle reminder that it's a oneof

`UpsertReporterStatus` is a helper for obtaining the value in the map (a `Status`) that corresponds to the `POD_NAMESPACE` of the current environment.  This one saves us from having repeated logic scattered across the codebase which calls `GetReporterStatus()`, then `os.Getenv("POD_NAMESPACE")` to determine the proper key to use to add/update a `Status` value in the `ReporterStatus.Statuses` map.

`GetStatus` is deprecated, because going forward it will be incorrect to use that function 99.9% of the time. The only time it would be valid is when we're reading a resource that has a `Status` that has not yet been written as a `ReporterStatus` (i.e. after upgrade).

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -87,8 +87,66 @@ func (r *{{ .Name }}) SetMetadata(meta *core.Metadata) {
...
+func (r *{{ .Name }}) HasReporterStatus() bool {
+	switch r.StatusOneof.(type) {
+	case *{{ .Name }}_ReporterStatus:
+		return true
+	default:
+		return false
+	}
+}
+
+func (r *{{ .Name }}) HasStatus() bool {
```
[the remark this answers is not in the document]
MitchAman: Ah, at the moment, a compilation failure.  I guess we're probably better off without these.

pkg/api/v1/clients/kube/crd/crd.go
```diff
@@ -106,14 +106,23 @@ func (d Crd) KubeResource(resource resources.InputResource) (*v1.Resource, error
...
+		if resource.HasStatus() {
 			statusProto := resource.GetStatus()
 			statusMap, err := protoutils.MarshalMapFromProtoWithEnumsAsInts(statusProto)
 			if err != nil {
 				return nil, MarshalErr(err, "resource status to map")
 			}
 			status = statusMap
 		}
+
+		if resource.HasReporterStatus() {
```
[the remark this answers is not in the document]
MitchAman: Good idea, done.

pkg/api/v1/clients/kube/resource_client.go
MitchAman: I like this idea, but not sure if there's a great way I can catch the specific error. For example, the error returned when attempting to unmarshal a `ReporterStatus` into a `Status` looks like: `unknown field "statuses" in core.solo.io.Status` of the type `*errors.errorString`.  Who knows if there are other unmarshalling errors that could be returned with a different message format?

If there was a `UnmarshalError` error type that was returned, it would be much easier to just look for that error type rather than trying to match a string from a generic error

pkg/api/v1/resources/resource_interface.go
sam-heilbron: Sorry I think my comment was a little misleading. I meant the GetNamespacedStatus and UpsertReporterStatus methods which deprecate the GetStatus, SetStatus methods, but share the same signature. I like the `Has` methods as well! 

I had imagined the interface for a resource would be Get/Set for both Status and ReporterStatus, but that the SetStatus implementation would change to write to the ReporterStatus first, and if not possible, fallback to the Status field, and likewise, the GetStatus implementation would change to read from the ReporterStatus field first, and fallback on the status field. 

Perhaps this isn't possible / doesn't make sense, though.

MitchAman: We need `GetStatus()` to return nil when the `status_oneof` contains a `NamespacedStatuses` (formerly ReporterStatus) so that we can detect what's in the oneof. HasStatus() and HasNamespacedStatuses() have been removed since the last comment

test/mocks/crds/testing.solo.io_v1_MockResource.yaml
MitchAman: I don't know a whole lot about these CRD schemas, but my guess is it's because status and namespacedStatuses are in a oneof.  So the 'status' keys in the schema now refer to a oneof value rather than a `Status` value.

sam-heilbron (commented):
I noticed when installing the gloo branch in a cluster, that upstreams used to take the following shape:

```
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  annotations:
    meta.helm.sh/release-name: gloo
    meta.helm.sh/release-namespace: gloo-system
  creationTimestamp: "2021-07-26T22:34:29Z"
  generation: 2
  labels:
    discovered_by: kubernetesplugin
  name: gloo-system-gateway-443
  namespace: gloo-system
  resourceVersion: "9699"
  uid: 7f306c90-78d6-40a2-9872-2a94e571cb49
spec:
  discoveryMetadata:
    labels:
      app: gloo
      app.kubernetes.io/managed-by: Helm
      discovery.solo.io/function_discovery: disabled
      gloo: gateway
  kube:
    selector:
      gloo: gateway
    serviceName: gateway
    serviceNamespace: gloo-system
    servicePort: 443
status:
  reportedBy: gloo
  state: 1
```

but now upstreams look like:

```
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  annotations:
    meta.helm.sh/release-name: gloo
    meta.helm.sh/release-namespace: gloo-system
  creationTimestamp: "2021-07-30T12:57:16Z"
  generation: 2
  labels:
    discovered_by: kubernetesplugin
  name: gloo-system-gloo-9976
  namespace: gloo-system
  resourceVersion: "1543"
  uid: 68a1040f-c476-48a2-a1c0-a4abb9f54b3a
spec:
  discoveryMetadata:
    labels:
      app: gloo
      app.kubernetes.io/managed-by: Helm
      gloo: gloo
  kube:
    selector:
      gloo: gloo
    serviceName: gloo
    serviceNamespace: gloo-system
    servicePort: 9976
  namespacedStatuses:
    statuses:
      gloo-system:
        reportedBy: gloo
        state: Accepted
status:
  statuses:
    gloo-system:
      reportedBy: gloo
      state: 1
```

It seems like we're now writing the namespacesStatuses to the spec as well as the status, on this InputResource, whereas previously we were not.

kdorosh:
superseded by https://github.com/solo-io/solo-kit/pull/437

PR #431, opened 2021-07-19
schema-gen: support description truncation and int64 primitive type

sam-heilbron:
# Description
- Fix issue with serialization of int64's and validation schemas: https://github.com/solo-io/gloo/issues/5043
- Expose options to control how large descriptions are in validation schemas: https://github.com/solo-io/gloo/issues/4789

# Context
## int64's in validation schemas (https://github.com/solo-io/gloo/issues/5043)

Int64's are serialized as strings. Since we introduced validation schemas, the schema expects an integer, but the value provided is a stringified version of an integer (ie "2"). A recent commit to the library we use for schema generation (https://github.com/sam-heilbron/protoc-gen-openapi/commit/e703b93862db646a58a3c5d0a83777e2c59b692d) fixed this, by including the `x-kubernetes-int-or-string` property (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#intorstring) on the schema.

### Testing
- [x] Introduced a unit test to verify that the generated schema for an `int64` field, includes the new `x-kubernetes-int-or-string` property.
- [x] Ran an end-to-end test to verify that this new field would behave as expected:
1. I pointed a local gloo branch, to the most recent solo-kit commit of this PR
1. I built the chart, and created a test cluster: `VERSION=v0.0.0-helm CLUSTER_NAME=helm-test CLUSTER_NODE_VERSION=v1.21.1 ./ci/deploy-to-kind-cluster.sh`
1. I installed gloo using that new chart: `helm install gloo _test/gloo-v0.0.0-helm.tgz --namespace gloo-system -f /path/to/yaml/file --create-namespace`, where path/to/yaml/file is the path to the values override file presented in the issue
1. I added a route so that a proxy would be generated, and verified that there was no error
1. I manually changed the validation schema on the proxy to remove the new `x-kubernetes-int-or-string` property, updated the route so the proxy would re-generate, and confirmed that an error now occurred.

## descriptions in validation schemas (https://github.com/solo-io/gloo/issues/4789)
Descriptions in validation schemas aren't functional, shouldn't be used as a source of truth around the API, and don't affect the error message presented when a resource errors due to validation schema constraints. Also, since descriptions are generated from the comments above the source proto field, and those comments also generate docs in Gloo, descriptions are quite verbose. This causes the CRD to be massive, containing lots of information that isn't relevant to the CRD itself. If the CRD get's too large, it can no longer be applied, which is what happened in this issue.
 
### Testing
- [x] Add tests to verify that descriptions can be either truncated or removed.


pkg/code-generator/schemagen/generator.go
```diff
@@ -22,6 +22,18 @@ type ValidationSchemaOptions struct {
 
 	// Tool used to generate JsonSchemas, defaults to protoc
 	JsonSchemaTool string
+
+	// Whether to remove descriptions from validation schemas
+	// Default: false
+	//
+	// NOTE: I'd prefer a positive field name (ie includeDescriptions)
+	//	but I wanted to avoid changing the default behavior
+	RemoveDescriptionsFromSchema bool
```
saiskee, line 31: [nit] What if `MaxDescriptionCharacters` was 0 and `RemoveDescriptionsFromSchema` is true? Will one override the other? Would be nice to have this in the comment to quickly grok the behavior.

go.mod
```diff
@@ -43,7 +43,7 @@ require (
 	github.com/pseudomuto/protokit v0.2.0
 	github.com/radovskyb/watcher v1.0.2
 	github.com/rotisserie/eris v0.1.1
-	github.com/sam-heilbron/protoc-gen-openapi v0.0.0-20210521134209-c323189e7a9a
+	github.com/sam-heilbron/protoc-gen-openapi v0.0.0-20210719212650-45522763336d
```
saiskee, line 46: I think this is fine for now, but if do any additional changes on prrotoc-gen-openapi, we should move it to the solo-io org and add changelog / release versioning to track breaking/other changes. Not super important right now ofc because this is the only repo that uses it.

sam-heilbron: Agreed!

pkg/code-generator/schemagen/generator.go
sam-heilbron: Yeah good call

saiskee (approved):

PR #434, opened 2021-07-23
Add keep pr updated label.

MitchAman:
Add the keep pr updated label to solo-kit.

### DIFF
--- a/.github/workflows/add-pr-label.yaml
+++ b/.github/workflows/add-pr-label.yaml
@@ -1,6 +1,8 @@
 name: PR Auto Labeling
 on:
-  pull_request: {}
+  pull_request:
+    types:
+      - opened
 jobs:
   add-label:
     name: Add `keep pr updated` Label
--- a/.github/workflows/add-pr-label.yaml
+++ b/.github/workflows/add-pr-label.yaml
@@ -1,8 +1,6 @@
 name: PR Auto Labeling
 on:
-  pull_request:
-    types:
-      - opened
+  pull_request: {}
 jobs:
   add-label:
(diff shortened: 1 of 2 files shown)


.github/workflows/add-pr-label.yaml
```diff
@@ -0,0 +1,17 @@
...
+    name: Add `keep pr updated` Label
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: christianvuerings/add-labels@v1
+        with:
+          labels: |
+            keep pr updated
+        env:
+          GITHUB_TOKEN: ${{ secrets.CI_ADMIN_GITHUB_TOKEN }}
```
kdorosh, line 17: try `PERSONAL_ACCESS_TOKEN`

.github/workflows/add-pr-label.yaml
```diff
@@ -0,0 +1,17 @@
+name: PR Auto Labeling
+on:
+  pull_request:
+    types:
+      - opened
```
kdorosh, line 5: we should change this temporarily to confirm that it works, then put it back to only trigger on opened pull request events (rather than any pr event, i.e. `pull_request: {}`)

MitchAman: Good call. Looks like it worked :)

.github/workflows/add-pr-label.yaml
MitchAman: This worked, thanks.

kdorosh (approved):

PR #437, opened 2021-08-19
Support Namespaced Statuses

sam-heilbron:
# Blocked
- [x] PR which introduced this into gloo is passing CI (https://github.com/solo-io/gloo/pull/5191)
- [x] PR which introduces this change into solo-projects is passing CI (https://github.com/solo-io/solo-projects/pull/2712) (*Only build-bot is failing on CI, and it's passing for me locally*)

# Description
Add support for having statuses from more than one controller. 

# Context
This PR adds a new proto message, NamespacedStatuses, that acts as a replacement for Status. NamespacedStatuses is a mapping from POD_NAMESPACE (env var) to Status. This allows multiple gloo instances in different namespaces to record their own statuses without overwriting one another.

This requires that controllers have the `POD_NAMESPACE` env variable set.

# https://github.com/solo-io/solo-kit/pull/429
Relies largely on the implementation of: https://github.com/solo-io/solo-kit/pull/429.

The main difference with this PR is that we do not introduce a oneof to support both fields. Instead, we handle the conversion when unmarshalling, and require that all projects which import this library, set the POD_NAMESPACE environment variable, for the controller that updates resources/statuses.


pkg/api/v2/reporter/reporter_suite_test.go
```diff
@@ -1,12 +1,29 @@
...
+	"github.com/solo-io/solo-kit/test/helpers"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+var namespace string
+
+var _ = BeforeSuite(func() {
+	namespace = helpers.RandString(5)
```
MitchAman, line 16: What's the purpose of the random string? I generally feel that tests should be as deterministic as possible to make errors more reproducible and minimize flakes

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -86,9 +87,63 @@ func (r *{{ .Name }}) SetMetadata(meta *core.Metadata) {
...
+	if podNamespace == "" {
+		return errors.NewPodNamespaceErr()
+	}
+	if r.GetNamespacedStatuses() == nil {
+		r.SetNamespacedStatuses(&core.NamespacedStatuses{})
+	}
+	if r.GetNamespacedStatuses().Statuses == nil {
+		r.GetNamespacedStatuses().Statuses = make(map[string]*core.Status)
+	}
+	r.GetNamespacedStatuses().Statuses[podNamespace] = status
```
MitchAman, line 124: `.Statuses` -> `.GetStatuses()`

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -86,9 +87,63 @@ func (r *{{ .Name }}) SetMetadata(meta *core.Metadata) {
...
+	if podNamespace == "" {
+		return nil, errors.NewPodNamespaceErr()
+	}
+	if r.GetNamespacedStatuses() == nil {
+		return nil, nil
+	}
+	if r.GetNamespacedStatuses().Statuses == nil {
+		return nil, nil
+	}
+	return r.GetNamespacedStatuses().Statuses[podNamespace], nil
```
MitchAman, line 144: `.Statuses` -> `.GetStatuses()`

pkg/code-generator/parser/parser_resource.go
```diff
@@ -227,7 +227,7 @@ func describeResource(messageWrapper ProtoMessageWrapper) (*model.Resource, erro
 	// always make it upper camel
 	pluralName = strcase.ToCamel(pluralName)
 
-	hasStatus := hasField(msg, "status", statusTypeName)
+	hasNamespacedStatuses := hasField(msg, "namespaced_statuses", namespacedStatusesTypeName)
```
MitchAman, line 230: Good call

test/mocks/tests/mock_resource_test.go
```diff
@@ -0,0 +1,159 @@
...
+			mockRes.SetNamespacedStatuses(namespacedStatuses)
+
+			Expect(mockRes.GetNamespacedStatuses()).To(BeEquivalentTo(namespacedStatuses))
+		})
+	})
+
+	Context("GetStatusForNamespace", func() {
+		It("Should return the correct status with respect to the POD_NAMESPACE", func() {
+			mockRes := v1.MockResource{}
+			ns1Status := Status{
```
MitchAman, line 49: nit: `ns1Status` -> `blueNamespaceStatus`, `ns2Status` -> `greenNamespaceStatus`,

pkg/api/v1/clients/kube/resource_client_test.go
```diff
@@ -48,7 +48,9 @@ var (
 	client     *kube.ResourceClient
 	clientset  *versioned.Clientset
 	lock       *clusterlock.TestClusterLocker
+	namespace  = helpers.RandString(5)
```
MitchAman, line 51: Same question about the randomness

pkg/api/v1/reconcile/reconcile_suite_test.go
```diff
@@ -1,12 +1,29 @@
...
+	"github.com/solo-io/solo-kit/test/helpers"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+var namespace string
+
+var _ = BeforeSuite(func() {
+	namespace = helpers.RandString(5)
```
MitchAman, line 16: ditto

pkg/api/v1/reporter/reporter_suite_test.go
```diff
@@ -1,12 +1,29 @@
...
+	"github.com/solo-io/solo-kit/test/helpers"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+var namespace string
+
+var _ = BeforeSuite(func() {
+	namespace = helpers.RandString(5)
```
MitchAman, line 16: ditto

pkg/api/v2/reporter/reporter_suite_test.go
sam-heilbron: Yeah good call, I had intially done this since other tests follow a similar pattern. But i don't think it's necessary for these tests so may as well keep it simple

pkg/api/v2/reporter/reporter.go
```diff
@@ -252,9 +262,9 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 			continue
 		}
 		if updatedResource != nil {
-			logger.Debugf("wrote report for %v : %v", updatedResource.GetMetadata().Ref(), status)
+			logger.Warnf("wrote report for %v : %v", updatedResource.GetMetadata().Ref(), status)
```
sam-heilbron, line 265: TODO - revert this change

MitchAman (commented):
Looking good! Were you able to confirm the status isn't showing up the spec as it was in my PR?

sam-heilbron:
> Looking good! Were you able to confirm the status isn't showing up the spec as it was in my PR?

Yeah I was able to verify this by running the following steps in Gloo (when pointing to my Gloo PR):
```
VERSION=0.0.1-namespaced-statuses CLUSTER_NAME=namespaced-statuses CLUSTER_NODE_VERSION=v1.21.1 ./ci/deploy-to-kind-cluster.sh
 helm install gloo-ns-statuses _test/gloo-0.0.1-namespaced-statuses.tgz --namespace gloo-system --create-namespace
```

I inspected the status on the upstreams that are generated during discovery and used `glooctl create vs --name vs-test` to create a virtual service. The glooctl check doesn't work yet, but when you edit the resource (or describe it) you see the appropriate yaml, without the duplicated status.

pkg/api/v2/reporter/reporter.go
MitchAman: ^

MitchAman (commented):
LGTM, aside from your comment above

MitchAman (dismissed):

kdorosh:
supersedes https://github.com/solo-io/solo-kit/pull/429

pkg/api/v1/clients/kube/crd/crd.go
```diff
@@ -103,16 +103,15 @@ func (d Crd) KubeResource(resource resources.InputResource) (*v1.Resource, error
 		}
 
 		delete(data, "metadata")
-		delete(data, "status")
+		delete(data, "namespacedStatuses")
```
kdorosh, line 106: k8s expects and may treat fields such as `metadata`, `spec`, and `status` in a special way. can we keep the field called `status` at the top level, but handle both namespaced and non-namespaced statuses under the covers while unmarshalling? e.g.

```
status:
  <status>
```

or

```
status:
  ns1:
    <status1>
  ns2:
    <status2>
```

rather than
```
namespacedStatuses:
  ns1:
    <status1>
  ns2:
    <status2>
```

pkg/code-generator/parser/parser_resource.go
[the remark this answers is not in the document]
sam-heilbron: I can add a note with some more details. This check is used by solo-kit to determine if a configured proto supports 'gloo status reporting'. Previously, the protos for the gloo API defined a `status` field with a particular type and now we will define a `namespaced_statuses` field with a new type. We could make this field whatever name we want, we just need to ensure that the name we use in gloo matches the expected name here

pkg/api/v1/clients/kube/crd/crd.go
sam-heilbron: Spoke offline (and will add comment to code):

We are still keeping the status as a top level field. What this line is doing is removing the status (now namespaced_statuses) field from the spec, so that when we persist it, it does not live both on the spec and the status.

kdorosh (commented):
lgtm, just a question

pkg/errors/errors.go
```diff
@@ -99,7 +97,7 @@ func IsResourceVersion(err error) bool {
 type podNamespaceErr struct{}
 
 func (err *podNamespaceErr) Error() string {
-	return fmt.Sprintf("%s environment variable is not set", envutils.PodNamespaceEnvName)
+	return fmt.Sprintf("%s environment variable is not set", "POD_NAMESPACE")
```
MitchAman, line 100: `statusutils.PodNamespaceEnvName` here?

MitchAman (commented):
Looks good - just one spot you may have missed?

pkg/errors/errors.go
sam-heilbron: I had tried this initially, but it was causing a circular dependency, so I opted to just re-define it as a string

MitchAman: Ah, gotcha.

sam-heilbron:
Closed in favor of: https://github.com/solo-io/solo-kit/pull/447

PR #438, opened 2021-08-19
Fix oneof docsgen

saiskee:
See https://github.com/solo-io/gloo/pull/5194/files#diff-e838e4f2e45d8e4ef164c9473b20980779cf31bb3deb96e1e8a02f96743248aaR213 for effect of fix.


test/mocks/tests/docsgen_test.go
```diff
@@ -0,0 +1 @@
+package tests
```
jenshu, line 1: why was this created?

jenshu (approved):

PR #441, opened 2021-08-21
Deflake ResourceClient CRUD tests

sam-heilbron:
## Background
We run solo-kit tests in parallel. To avoid having kubernetes based tests step on eachothers toes, we have each test run in a separate/randomized namespace. 

## What is the current race condition/ test pollution?
Our kubernetes clients tests share a test utility that is provided unique namespaces, and performs a series of client actions. However, one of these actions was to list resources in the `""` namespace, which is treated as "all namespaces". The purpose of the test is to validate that a call to client.List will return an empty list instead of an error. However, since it lists all resources in all namespaces, if other tests are running in parallel, there is a race condition where it can find those resources being used by other tests.

## How did this error present itself?
https://console.cloud.google.com/cloud-build/builds/9b97abd8-1e76-4598-bcbc-f05225806624;step=7?project=solo-public

This test has failed every now and then, but recently it picked up in frequency. We had a series of releases and PR's fail with the same error.

```
• Failure [7.384 seconds]
PlainConfigmap
/workspace/solo-kit/pkg/api/v1/clients/configmap/tests/plain/plain_resource_client_test.go:28
  CRUDs resources [It]
  /workspace/solo-kit/pkg/api/v1/clients/configmap/tests/plain/plain_resource_client_test.go:59

  Expected
      <resources.ResourceList | len:2, cap:2>: [ .... {INSERT RESOURCE CONTENT} ]
  to be empty

  /workspace/solo-kit/test/tests/generic/test_crud_client.go:38
```

## Solution
Instead of having the test list resources in all namespaces, we have it list the resources in the injected namespace. It provides the same validation (that an empty list is returned instead of an error) and ensures that as long as we provide unique namespaces to this test utility, that we do not have a race condition/test pollution in our testing.
 
## How did I verify this?
- [x] Updated the `test` make target to include `untilItFails` and then run:` RUN_KUBE_TESTS=1 WAIT_ON_FAIL=1 make test`. 
- [x] The test passed CI once, I kicked them again to verify they passed (https://github.com/solo-io/solo-kit/pull/441#issuecomment-903164387)

### DIFF
--- a/pkg/api/v1/clients/file/resource_client_test.go
+++ b/pkg/api/v1/clients/file/resource_client_test.go
@@ -3,6 +3,7 @@ package file_test
 import (
 	"context"
 	"os"
+	"path/filepath"
 	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


test/tests/generic/test_crud_client.go
```diff
@@ -30,32 +32,37 @@ func TestCrudClient(namespace1, namespace2 string, client ResourceClient, opts c
 	input.Metadata.Labels = labels
 
 	err := client.Register()
-	Expect(err).NotTo(HaveOccurred())
+	ExpectWithOffset(testOffset, err).NotTo(HaveOccurred())
+
+	// list with no resources should return empty list, not err
+	list, err := client.List(namespace1, clients.ListOpts{})
```
sam-heilbron, line 38: This is the functional change of this PR. Before we listed resources in the "" namespace. Now we make subsequent calls to the specific namespace.

test/tests/generic/test_crud_client.go
```diff
@@ -17,9 +17,11 @@ import (
 
 // Call within "It"
 func TestCrudClient(namespace1, namespace2 string, client ResourceClient, opts clients.WatchOpts, callbacks ...Callback) {
+	testOffset := 1
```
sam-heilbron, line 20: Since this test is called within in `It`, we need to use ExpectWithOffset(1) so that errors are accurately displayed

pkg/api/v1/clients/file/resource_client_test.go
```diff
@@ -18,24 +19,40 @@ import (
...
+		namespace2 = "ns2"
 	)
+
 	JustBeforeEach(func() {
 		var err error
+
 		tmpDir, err = ioutil.TempDir("", "base_test")
 		Expect(err).NotTo(HaveOccurred())
+
+		// Create a directory per namespace inside the tmpDir.
```
sam-heilbron, line 36: Because of the change we made below, to list resources in a specific namespace, instead of all namespaces, must must ensure that the namespace directories are created during the setup of the file resource client test.

sam-heilbron:
Tests passed: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=8cf8d33a-f46c-43ae-aa9a-127351984f94

Kicking again to be certain

/kick

saiskee (approved):

PR #451, opened 2021-09-23
Generate v1 CRDs

elcasteel:
Generate CRDs in the v1 format. The changes between API versions are documented here: https://kubernetes.io/docs/reference/using-api/deprecation-guide/ 
The biggest change for our CRDs was moving to a list of versions.

Also update ResourceClients and tests to use the newer format.


pkg/code-generator/schemagen/generator.go
```diff
@@ -85,60 +85,62 @@ func GenerateOpenApiValidationSchemas(project *model.Project, options *Validatio
...
-		}
 
-		validationSchema.OpenAPIV3Schema.Properties["spec"] = *specJsonSchema
-		validationSchema.OpenAPIV3Schema.Properties["status"] = statusSchema
+			validationSchema.OpenAPIV3Schema.Properties["spec"] = *specJsonSchema
+			validationSchema.OpenAPIV3Schema.Properties["status"] = statusSchema
 
-		if err = crdWriter.ApplyValidationSchemaToCRD(crd, validationSchema); err != nil {
-			return err
+			if err = crdWriter.ApplyValidationSchemaToCRD(crd, validationSchema); err != nil {
```
sam-heilbron, line 130: ApplyValidationSchemaToCRD was probably a poorly named function by me in the first place, but this will write the validation schema for a particualr CRD to a file. Right now we iterate over each version (even though we only support 1) and always write that to the 1st validation schema. Therefore, if we supported multiple versions, they would each be overwritten. 

Can we either:
(A) explicitly only support 1 and throw an error if there are more than 1
or 
(B) change how we call the ApplyValidationSchemaToCRD to support validation schemas per version

pkg/code-generator/schemagen/crd.go
```diff
@@ -62,12 +56,12 @@ func (c *CrdWriter) ApplyValidationSchemaToCRD(crd apiextv1beta1.CustomResourceD
 	})
 }
 
-func getFilenameForCRD(crd apiextv1beta1.CustomResourceDefinition) string {
-	return fmt.Sprintf("%s_%s_%s.yaml", crd.Spec.Group, crd.Spec.Version, crd.Spec.Names.Kind)
+func getFilenameForCRD(crd apiextv1.CustomResourceDefinition) string {
+	return fmt.Sprintf("%s_%s_%s.yaml", crd.Spec.Group, crd.Spec.Versions[0].Name, crd.Spec.Names.Kind)
```
sam-heilbron, line 60: now that we support (or will support) multple versions of the CRD. Does it make sense to make this file name just be ${GROUP}_${KIND}?

elcasteel: Since we decided to not support multiple versions yet, I think we should keep the name of the version we used in the filename.

sam-heilbron (approved):

PR #453, opened 2021-12-01
fix(code-generator): code generation now copies all *.go files

gunnar-solo:
related https://github.com/solo-io/gloo/issues/5620


nfuden:
The changelog entry does not match whats going on in the code change.

Changelog implies that the change should be from hardcoded -> string matcher but its going from a set of string matchers to a singular one.

elcasteel:
Can you also provide context for why we are copying all *.go files?

gunnar-solo:
sure, @elcasteel!  This change _could_ have been accomplished by adding a new case for `.pb.clone.go`.  _But_, that would effectively kick the problem down the line for the next person.

The temporary folder that generated files go to is populated like this:
Screen Shot 2021-12-01 at 2 35 35 PM

The thought process behind this PR:  making a less-restrictive copy operation is likely going to save work the next time something comes up similar to this (rather than coding yet another specifically-named case)

EItanya (approved):

PR #462, opened 2022-04-09
Update `Snapshot` interface to have `MakeConsistent` function

kdorosh:
For context behind these changes, see the PR description on https://github.com/solo-io/gloo/pull/6289

### DIFF
--- a/pkg/api/v1/control-plane/cache/generic_snapshot.go
+++ b/pkg/api/v1/control-plane/cache/generic_snapshot.go
@@ -21,6 +21,11 @@ import (
 	"github.com/golang/protobuf/proto"
 )
 
+var (
+	// Compile-time assertion
+	_ Snapshot = new(GenericSnapshot)
+)
+
 type TypedResources map[string]Resources
 
 type GenericSnapshot struct {
--- a/pkg/api/v1/control-plane/cache/generic_snapshot.go
+++ b/pkg/api/v1/control-plane/cache/generic_snapshot.go
@@ -15,10 +15,12 @@
 package cache
 
 import (
+	"context"
 	"errors"
 	"fmt"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/solo-io/go-utils/contextutils"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


kdorosh:
/kick looks like flake https://console.cloud.google.com/cloud-build/builds;region=global/d6ecb283-5e89-4bb6-b884-15bce8759ae6;step=7?project=solo-public

pkg/api/v1/control-plane/cache/generic_snapshot.go
```diff
@@ -103,17 +110,26 @@ func (s *GenericSnapshot) Consistent() error {
...
+				return fmt.Errorf("required resource name not in snapshot: %s %s", ref.Type, ref.Name)
 			}
 		} else {
-			return fmt.Errorf("required resource not in snapshot: %s %s", ref.Type, ref.Name)
+			return fmt.Errorf("required resource type not in snapshot: %s %s", ref.Type, ref.Name)
 		}
 	}
 
 	return nil
 }
```
nfuden, line 123:
```suggestion
// MakeConsistent  should never be called on a generic snapshot as it is not used for snapshots with dependent resources.
```

pkg/api/v1/control-plane/cache/generic_snapshot.go
```diff
@@ -103,17 +110,26 @@ func (s *GenericSnapshot) Consistent() error {
...
 
+func (s *GenericSnapshot) MakeConsistent() {
+	// this is fine since generic snapshots are only used by extauth/ratelimit extensions syncers; and those don't
+	// have dependent resources. this will not be called anywhere
+	contextutils.LoggerFrom(context.TODO()).DPanicf("it is an error to call make consistent on a generic snapshot")
```
nfuden, line 127: nit: Do we really want to panic? Unless we are catching the panic it blows stuff up and in general we should avoid em.
Perhaps since this truly should never happen its ok to crash out to the top level shutdown loop for a service but I just wanted to make sure we considered this.

nfuden (commented):
While I still wonder about the general removal of errors being returned as values I think overall the changes make sense and the additional comments do provide guidance as to future steps which is great.

I had 2 small questions that fall under the nit category here I am fine letting it through without them because bulldozer exists I am holding off on the approval just so that the 2 questions can be looked at.

pkg/api/v1/control-plane/cache/generic_snapshot.go
kdorosh: added comments for more clarity in https://github.com/solo-io/solo-kit/pull/462/commits/0859c25550a4c94a34e724db0b9f679805ef03bd

nfuden (approved):

PR #463, opened 2022-04-11
Support custom Vault secret prefix

inFocus7:
Updates
- Add `PathPrefix` `VaultSecretClientFactory` & clientResource.
- Update `resourceDirectory` to use new pathPrefix (or default to "secret/"), instead of hard-coding to `secret`.
- Create `DefaultTestPathPrefix` const for tests to use
- Add path prefix to tests + update `vault secrets` test setup command to use the option `-path=<pathPrefix>`.

### DIFF
--- a/pkg/api/v1/clients/vault/resource_client.go
+++ b/pkg/api/v1/clients/vault/resource_client.go
@@ -312,7 +312,7 @@ func (rc *ResourceClient) resourceDirectory(namespace, directoryType string) str
 	}
 
 	return strings.Join([]string{
-		"secreto",
+		rc.pathPrefix,
 		directoryType,
 		rc.root,
 		rc.resourceType.GroupVersionKind().Group,
--- a/pkg/api/v1/clients/vault/resource_client.go
+++ b/pkg/api/v1/clients/vault/resource_client.go
@@ -70,13 +70,15 @@ func (rc *ResourceClient) toVaultSecret(resource resources.Resource) (map[string
 type ResourceClient struct {
 	vault        *api.Client
 	root         string
+	pathPrefix   string
 	resourceType resources.VersionedResource
 }
 
-func NewResourceClient(client *api.Client, rootKey string, resourceType resources.VersionedResource) *ResourceClient {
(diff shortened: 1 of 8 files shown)


pkg/api/v1/clients/vault/resource_client.go
```diff
@@ -305,8 +307,12 @@ const (
 )
 
 func (rc *ResourceClient) resourceDirectory(namespace, directoryType string) string {
+	if rc.pathPrefix == "" {
```
sam-heilbron, line 310: nit: I think it makes it easier to reason through if this default behavior is defined in the constructor. 

Also, the rootKey does not default in solo-kit and instead we accept whatever value is injected. Is there a particular reason we don't follow a similar pattern?

sam-heilbron (commented):
Looking good! Some small comments (non-blocking) around ways that I think we can make this code easier to reason through for future devs.

pkg/api/v1/clients/vault/resource_client.go
```diff
@@ -70,13 +70,15 @@ func (rc *ResourceClient) toVaultSecret(resource resources.Resource) (map[string
 type ResourceClient struct {
 	vault        *api.Client
 	root         string
+	pathPrefix   string
 	resourceType resources.VersionedResource
 }
 
-func NewResourceClient(client *api.Client, rootKey string, resourceType resources.VersionedResource) *ResourceClient {
+func NewResourceClient(client *api.Client, rootKey string, pathPrefix string, resourceType resources.VersionedResource) *ResourceClient {
```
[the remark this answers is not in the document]
inFocus7: True, thanks! I'll update the changelog for a new major version. I'll also probably move `pathPrefix` before `rootKey`, since that's the way the `resourceDirectory` the client works with is created.

pkg/api/v1/clients/vault/resource_client.go
inFocus7: Good callout. I was thinking of ensuring it at least defaults to `secret` since, in vault that's the behaviour when creating new clients. So if someone where to create a client connecting to a regular kv vault, they wouldn't have to specify it being "secret".

sam-heilbron (approved):
Well done!

PR #464, opened 2022-04-19
Consul consistency mode + query options configuration

inFocus7:
- Added `queryOptions` to the factory (and resource client) set custom Consul query options.
  - Went this route instead of passing in `AllowStale` to the rc client. I did so to make it easier to configure/add to this in the future. We'd just have to add the new option we're adding into the `api.QueryOption{}` we're passing when creating an new client factory in Gloo.


jackstine (approved):
LGTM

PR #467, opened 2022-05-05
Use solo.io namespaced package for discovery.proto

birkland:
`api/external/envoy/api/v2/discovery.proto` is a fork of an envoy API published in [envoyproxy/data-plane-api](https://github.com/envoyproxy/data-plane-api/blob/main/envoy/api/v2/discovery.proto).  Unlike other such forked .protos in solo-kit, its package is not prefixed with `solo.io`.   Generated code from solo kit under `package envoy.api.v2` will conflict with generated code published in [envoyproxy/go-control-plane](https://github.com/envoyproxy/go-control-plane/blob/main/envoy/api/v2/discovery.pb.go), for any project that has both as a dependency.  Unfortunately, gloo-mesh is one such project.

This PR brings `discovery.proto` into alignment with the other forked envoy protos in solo-kit by namespacing the package to `solo.io`.  

related to solution for: https://github.com/solo-io/gloo-mesh-enterprise/issues/3624


birkland:
closing because this change was rolled back previously, as it breaks wire compatibility: https://github.com/solo-io/gloo/pull/4777#discussion_r643980041

PR #468, opened 2022-05-16
Support set-based label selectors

sam-heilbron:
# Description
Add support for resource clients to select objections with labels using an expression

# Context
The label selector is the core grouping primitive in Kubernetes. In solo-kit we previously only supported selecting resources from the apiserver via [equality-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#equality-based-requirement). 

## Why is this helpful?
Gloo clients should support the full feature-set of the Kubernetes API. 

Additionally, when Gloo controllers make changes to labels, we need to be able to support the intermediary period where we are looking for labels with multiple values.

## Is this backwards compatible?
Yes. This PR exposes the injection of an "ExpressionSelector". It is the preferred mechanism for defining selection, since it is the more powerful of the two. However, if it is not defined (as will be the case for all clients that upgrade to this version) we fallback to the original implementation which relies on the "Selector".

## How is this tested?
- [x] There are e2e tests defined, which create resources with varying labels, and then use resource clients to list those resources and validate the objects returned.
- [x]  There are e2e tests defined, which create resources with varying labels, and then use resource clients to list those resources and validate the objects returned.
- [x] There is a Gloo PR which incorporates these changes and proves it works as expected.  This isn't entirely necessary, since the changes can/should be able to validated within this library. https://github.com/solo-io/gloo/pull/6461 I was able to run the new regression test locally as well.

### DIFF
--- a/pkg/api/v1/clients/client_interface.go
+++ b/pkg/api/v1/clients/client_interface.go
@@ -104,9 +104,10 @@ func (o DeleteOpts) WithDefaults() DeleteOpts {
 }
 
 type ListOpts struct {
-	Ctx      context.Context
-	Selector map[string]string
-	Cluster  string
+	Ctx                context.Context
+	Selector           map[string]string
+	ExpressionSelector string
+	Cluster            string
 }
 
 func (o ListOpts) WithDefaults() ListOpts {
@@ -120,9 +121,10 @@ func (o ListOpts) WithDefaults() ListOpts {
 // To achieve a similar behavior you can use the KubeResourceClientFactory.ResyncPeriod field. The difference is that it
 // will apply to all the watches started by clients built with the factory.
 type WatchOpts struct {
-	Ctx         context.Context
-	Selector    map[string]string
-	RefreshRate time.Duration
+	Ctx                context.Context
+	Selector           map[string]string
+	ExpressionSelector string
+	RefreshRate        time.Duration
 	// Cluster is ignored by aggregated watches, but is respected by multi cluster clients.
 	Cluster string
 }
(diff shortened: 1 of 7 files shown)


pkg/api/v1/clients/client_interface.go
```diff
@@ -120,9 +139,28 @@ func (o ListOpts) WithDefaults() ListOpts {
...
+	Selector map[string]string
+	// Set-based label requirements
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement
+	// Set-based label requirements allow filtering keys according to a set of values.
+	// Three kinds of operators are supported: in,notin and exists (only the key identifier).
+	// Example:
+	//	"product in (edge, mesh)" would return all objects with the label key equal to product
+	//	and value equal to edge or mesh
+	// If both ExpressionSelector and Selector are defined, ExpressionSelector is preferred
+	ExpressionSelector string
```
inFocus7, line 162: while this change is made to _allow_ set-based requirements, does this field act as a generic label selector string, where users can write any kind of selection they want? 👀 If so, not sure if it'd be worth modifying (someday) the description of this, since ~to me~ it implies it's just for set-based selections.

ex:
`ExpressionSelector: partition in (customerA, customerB),environment!=qa`
 where we could use both set-based and equality based.
or, instead of `Selector: [a: b, c: d]`, we can `ExpressionSelector: a=b,c=d`

inFocus7 (commented):
left a single comment/q but lgtm as well! 👍

pkg/api/v1/clients/client_interface.go
sam-heilbron: I like the suggestion. https://github.com/solo-io/solo-kit/pull/469 aims to resolve this.

PR #469, opened 2022-05-16
Update comment around ListOpts and WatchOpts

sam-heilbron:
Follow up to resolve https://github.com/solo-io/solo-kit/pull/468#discussion_r874067549


inFocus7:
Could not adding a changelog possibly mess with the code releases, since this change is in between/part of code added/to-be-released in `v0.27.1`? 👀

sam-heilbron:
> Could not adding a changelog possibly mess with the code releases, since this change is in between/part of code added/to-be-released in `v0.27.1`? 👀

I don't thing this will be an issue. Since v0.27.1 hasn't been released, it's the next one to be released. When that release is kicked off we will search for the v0.27.1 folder, and since there is one (from the previous PR) we will generate the notes from that content. It's a good callout however, that if we had an empty changelog, we would need to add one so that there is at least one in the folder.

PR #471, opened 2022-05-17
[Gloo 1.9, 1.10] Support set-based label selectors

sam-heilbron:
Backport of: https://github.com/solo-io/solo-kit/pull/468, https://github.com/solo-io/solo-kit/pull/469

Gloo 1.10 relies on [v0.24.2](https://github.com/solo-io/gloo/blob/v1.10.x/go.mod#L74), and [1.9 as well](https://github.com/solo-io/gloo/blob/v1.9.x/go.mod#L89). To avoid pulling in potentially breaking changes into that branch, I opted to backport this feature. I created a branch off of the `v0.24.2` tag, following the naming convention for our branch `v.0.[N].x`

# Description
Add support for resource clients to select objections with labels using an expression

# Context
The label selector is the core grouping primitive in Kubernetes. In solo-kit we previously only supported selecting resources from the apiserver via [equality-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#equality-based-requirement). 

## Why is this helpful?
Gloo clients should support the full feature-set of the Kubernetes API. 

Additionally, when Gloo controllers make changes to labels, we need to be able to support the intermediary period where we are looking for labels with multiple values.

## Is this backwards compatible?
Yes. This PR exposes the injection of an "ExpressionSelector". It is the preferred mechanism for defining selection, since it is the more powerful of the two. However, if it is not defined (as will be the case for all clients that upgrade to this version) we fallback to the original implementation which relies on the "Selector".

## How is this tested?
- [x] There are e2e tests defined, which create resources with varying labels, and then use resource clients to list those resources and validate the objects returned.
- [x]  There are e2e tests defined, which create resources with varying labels, and then use resource clients to list those resources and validate the objects returned.


inFocus7 (approved):
LGTM! 🚢

PR #472, opened 2022-05-23
Status field with supergraph for gateway schemas

Bslabe123:
BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/3338


pkg/api/v2/reporter/reporter.go
```diff
@@ -16,9 +16,18 @@ import (
...
+}
+
+type Snapshots struct {
+	graphqlSnapshot GraphqlSnapshot
+}
+
 type Report struct {
-	Warnings []string
-	Errors   error
+	snapshots Snapshots
```
kdorosh, line 28: let's add these fields instead https://github.com/solo-io/gloo-mesh-enterprise/blob/284fa6ad390c024088549bed26c13cadfea3e348/pkg/api/common.gloo.solo.io/v2/status.pb.go#L346-L356

pkg/api/v2/reporter/reporter.go
```diff
@@ -284,6 +293,8 @@ func (r *reporter) attemptUpdateStatus(ctx context.Context, client ReporterResou
 		if inputResourceFromRead, ok := resourceFromRead.(resources.InputResource); ok {
 			resourceToWrite = inputResourceFromRead
 			r.statusClient.SetStatus(resourceToWrite, statusToWrite)
```
kdorosh, line 295: we probably want a new api beyond `r.statusClient.SetStatus` that also allows us to set the new fields we're adding e.g. `SetMessage` and `SetUpdatedAt`; or preferably a well named function that has us set all the fields at once

pkg/api/v1/reporter/reporter.go
```diff
@@ -46,7 +46,7 @@ func (e ResourceErrors) Validate() error {
 }
 
 type Reporter interface {
-	WriteReports(ctx context.Context, errs ResourceErrors, subresourceStatuses map[string]*core.Status) error
+	WriteReports(ctx context.Context, errs ResourceErrors, subresourceStatuses map[string]*core.Status, messages []string) error
```
kdorosh, line 49: in our repos where the users are other repos (i.e. this is a library for another repo) we mark changes to public APIs such as this as breaking changes, since consumers (other code) will have to update so it compiles. thus the changelog here should be `BREAKING_CHANGE` and the changelog version should be updated accordingly

note that we don't do the same for Gloo / Solo-Projects (in those repos this would be non-breaking) since this would require a major version bump, something we only do once per quarter. Also, our users there are mostly people interacting with our CRs/yaml (which we should _never_ break), so this change is non-breaking to them

pkg/api/v2/reporter/reporter.go
```diff
@@ -19,6 +22,11 @@ import (
 type Report struct {
 	Warnings []string
 	Errors   error
+
+	// Additional information about the current state of the resource.
+	Messages []string `protobuf:"bytes,3,opt,name=message,proto3" json:"message,omitempty"`
+	// The last time the status was updated.
+	UpdatedAt *timestamp.Timestamp `protobuf:"bytes,4,opt,name=updated_at,json=updatedAt,proto3" json:"updated_at,omitempty"`
```
kdorosh, line 29: nit, let's remove the proto struct tags, since these are not protos. we can keep the json ones per https://gobyexample.com/json

pkg/api/v2/reporter/reporter_test.go
```diff
@@ -20,19 +20,20 @@ import (
 	v1 "github.com/solo-io/solo-kit/test/mocks/v1"
 )
 
-var _ = Describe("Reporter", func() {
+var _ = FDescribe("Reporter", func() {
```
kdorosh, line 23: we will need to unfocus for ci

pkg/api/v2/reporter/reporter.go
kdorosh: also i don't believe `UpdatedAt` is used anywhere

pkg/api/v2/reporter/reporter.go
```diff
@@ -195,20 +203,22 @@ type StatusReporter interface {
 }
 
 type reporter struct {
-	reporterRef  string
-	statusClient resources.StatusClient
-	clients      map[string]ReporterResourceClient
+	reporterRef    string
+	statusClient   resources.StatusClient
+	messagesClient resources.MessagesClient
```
kdorosh, line 208: since the messages live on the k8s status resource, i think we should have a single `statusClient` that happens to write the status we have today as well as any messages / additional info

pkg/api/v1/reporter/reporter.go
Bslabe123: Actually this is something i did incorrectly that I just rolled back but good to know for when it does come up!

pkg/api/v2/reporter/reporter.go
kdorosh: see https://github.com/solo-io/solo-kit/pull/472/files#r882880743 for how we might want to implement these new functions (on a status client, not a new client)

pkg/api/v2/reporter/reporter.go
Bslabe123: I meant to ask you about that, is this something we want to support or should it be removed.

pkg/api/v2/reporter/reporter_test.go
Bslabe123: Just fixed!

pkg/api/v2/reporter/reporter.go
Bslabe123: That does make the code alot less messy too, I can get that fixed up!

changelog/v0.28.0/supergraph-on-status.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
```
kdorosh, line 2:
```suggestion
  - type: BREAKING_CHANGE
```

kdorosh: i would read https://github.com/solo-io/go-utils/tree/master/changelogutils#changelog

pkg/api/v1/resources/resource_interface.go
```diff
@@ -51,6 +51,8 @@ type InputResource interface {
 	SetStatus(status *core.Status)
 	GetNamespacedStatuses() *core.NamespacedStatuses
 	SetNamespacedStatuses(namespacedStatuses *core.NamespacedStatuses)
```
kdorosh, line 53: can we perhaps embed the messages in the pre-existing details?

```proto
message Status {
    enum State {
        // Pending status indicates the resource has not yet been validated
        Pending = 0;
        // Accepted indicates the resource has been validated
        Accepted = 1;
        // Rejected indicates an invalid configuration by the user
        // Rejected resources may be propagated to the xDS server depending on their severity
        Rejected = 2;
        // Warning indicates a partially invalid configuration by the user
        // Resources with Warnings may be partially accepted by a controller, depending on the implementation
        Warning = 3;
    }
    // State is the enum indicating the state of the resource
    State state = 1;
    // Reason is a description of the error for Rejected resources. If the resource is pending or accepted, this field will be empty
    string reason = 2;
    // Reference to the reporter who wrote this status
    string reported_by = 3;
    // Reference to statuses (by resource-ref string: "Kind.Namespace.Name") of subresources of the parent resource
    map<string, Status> subresource_statuses = 4;

    // Opaque details about status results
    google.protobuf.Struct details = 5;
}
```

a namespaced status already supports this opaque metadata

this is better since it will play nicely with our multi-controller status reporter, rather than requiring you to reimplement

changelog/v0.27.3/supergraph-on-status.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/solo-projects/issues/3338
+    description: Add graphqlSnapshot field to status to Report for exposing computed supergraph
```
kdorosh, line 4: nit, this description is now dated

pkg/api/v1/resources/resource_interface_test.go
```diff
@@ -65,6 +65,16 @@ func (m *mockResources) SetStatus(status *core.Status) {
...
+func (m *mockResources) GetMessages() []string {
+	// Not needed in this test
+	panic("implement me")
+}
+
+func (m *mockResources) SetMessages(messages []string) {
+	// Not needed in this test
+	panic("implement me")
+}
+
```
kdorosh, line 77: no longer needed

pkg/api/v2/reporter/reporter.go
```diff
@@ -19,6 +19,9 @@ import (
 type Report struct {
 	Warnings []string
 	Errors   error
+
+	// Additional information about the current state of the resource.
+	Messages []string `protobuf:"bytes,3,opt,name=message,proto3" json:"message,omitempty"`
```
kdorosh, line 24: don't need the proto tags

pkg/api/v2/reporter/reporter.go
```diff
@@ -266,17 +269,17 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 	return merr.ErrorOrNil()
 }
 
-// Ideally, this and its caller, WriteReports, would just take the resource ref and its status, rather than the resource itself,
-//    to avoid confusion about whether this may update the resource rather than just its status.
+// Ideally, this and its caller, WriteReports, would just take the resource ref and its status and/or messages, rather than the resource itself,
+//    to avoid confusion about whether this may update the resource rather than just its fields.
 //    However, this change is not worth the effort and risk right now. (Ariana, June 2020)
-func (r *reporter) attemptUpdateStatus(ctx context.Context, client ReporterResourceClient, resourceToWrite resources.InputResource, statusToWrite *core.Status) (resources.Resource, resources.InputResource, error) {
+func (r *reporter) attemptUpdate(ctx context.Context, client ReporterResourceClient, resourceToWrite resources.InputResource, statusToWrite *core.Status, messagesToWrite []string) (resources.Resource, resources.InputResource, error) {
```
kdorosh, line 275: no longer need messagesToWrite? as messages is now a part of `core.Status`?

pkg/api/v2/reporter/reporter.go
```diff
@@ -246,7 +249,7 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 		var updatedResource resources.Resource
 		writeErr := errors.RetryOnConflict(retry.DefaultBackoff, func() error {
 			var writeErr error
-			updatedResource, resourceToWrite, writeErr = r.attemptUpdateStatus(ctx, client, resourceToWrite, status)
+			updatedResource, resourceToWrite, writeErr = r.attemptUpdate(ctx, client, resourceToWrite, status, report.Messages)
```
kdorosh, line 252: why change the name here?

pkg/api/v2/reporter/reporter.go
```diff
@@ -128,6 +131,21 @@ func (e ResourceReports) AddWarning(res resources.InputResource, warning string)
...
+	for _, warn := range message {
+		e.AddMessages(res, warn)
+	}
+}
+
+func (e ResourceReports) AddMessage(res resources.InputResource, message string) {
+	if message == "" {
+		return
+	}
+	rpt := e[res]
```
kdorosh, line 144: check if not found, and return if `res` is not in the map

pkg/api/v2/reporter/reporter.go
```diff
@@ -128,6 +131,21 @@ func (e ResourceReports) AddWarning(res resources.InputResource, warning string)
...
+	rpt.Messages = append(rpt.Warnings, message)
+	e[res] = rpt
```
kdorosh, line 146: do we really need the extra copy here? we should be able to append the message to the list in-place

pkg/api/v2/reporter/reporter.go
```diff
@@ -233,10 +251,10 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
 		if !ok {
 			return errors.Errorf("reporter: was passed resource of kind %v but no client to support it", kind)
 		}
-		status := r.StatusFromReport(report, subresourceStatuses)
+		status := r.StatusFromReport(report, subresourceStatuses) //HERE
```
kdorosh, line 254: remove //HERE

pkg/api/v2/reporter/reporter_test.go
kdorosh: bump

pkg/api/v2/reporter/reporter.go
Bslabe123: Fixed!

pkg/api/v2/reporter/reporter.go
Bslabe123: Added!

pkg/api/v2/reporter/reporter.go
Bslabe123: Removed

pkg/api/v2/reporter/reporter.go
Bslabe123: Fixed

changelog/v0.28.0/supergraph-on-status.yaml
Bslabe123: Sticking with NEW_FEATURE because the breaking change was rolled back if thats alright

pkg/api/v1/resources/resource_interface.go
Bslabe123: Added!

pkg/api/v1/resources/resource_interface_test.go
Bslabe123: Fixed!

pkg/api/v2/reporter/reporter.go
Bslabe123: This was an old change that wasnt rolled back, fixed

pkg/api/v2/reporter/reporter.go
Bslabe123: Same as above

pkg/api/v2/reporter/reporter.go
Bslabe123: Rmoved those

pkg/api/v2/reporter/reporter.go
Bslabe123: But then we wouldnt be able to create new messages, only modify existing ones wouldnt we?

pkg/api/v2/reporter/reporter.go
```diff
@@ -53,6 +56,17 @@ func (e ResourceReports) Merge(resErrs ResourceReports) {
 				}
 			}
 
+			// Merge messages lists
+			allMessages := make(map[string]bool)
```
kdorosh, line 60: nit; style. let's use a map[string]struct{} whenever we don't need the value and are using a map to implement a set

pkg/api/v2/reporter/reporter.go
kdorosh: what happens if `e[res]` returns `nil, false`? i.e. it's not in the map.. then on line 156 I think this will panic since `rpt.Messages` will dereference a nil pointer. Let's add a unit test to prove this out / catch this

pkg/api/v2/reporter/reporter.go
Bslabe123: Like last time, got it!

pkg/api/v2/reporter/reporter.go
```diff
@@ -314,6 +343,10 @@ func (r *reporter) attemptUpdateStatus(ctx context.Context, client ReporterResou
 }
 
 func (r *reporter) StatusFromReport(report Report, subresourceStatuses map[string]*core.Status) *core.Status {
+	var messages []string
+	if len(report.Messages) != 0 {
+		messages = report.Messages
+	}
```
kdorosh, line 349: how is this different than just always passing report.Messages? https://go.dev/play/p/4k0jGHToZC4

Bslabe123: This is just a check so that messages will always be nil rather than an an empty array. Figured this might be good futureproofing

PR #474, opened 2022-06-06
ordering-struct-for-watches

jackstine:
Updated for go1.16. 
If we want to wait for go1.18 then we can just swap out the stucts from the utils.

If you have questions about the struct, please ask them in the [Priority Struct PR](https://github.com/solo-io/go-utils/pull/483).  Please only discuss `solo-kit` changes here. 

Using the PrioirtyStruct to order the watches by priority. This structure is used to add and delete watches in optimal time.
The primary priorities that matter are `Cluster` and `Listener`.  

The issue is that currently there can be unordered updates by priority. `ClusterLoadAssignments` can be updated before `Cluster`s are updated. This causes a timeout issue in Envoy, resulting in all Endpoints on the cluster to be deleted, because the `ClusterLoadAssignement` update occurred before the `Cluster` update.

Not sure if we want to move the types into its own package. Can't use them because of circular dependency.


pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
...
+// If a priority does not exist for an added element, it is added to the lowest priority.
+type PrioritySortedStruct struct {
+	// prioritySets is the set lists for priorities, where P is the type used for priority.
+	prioritySets map[int][]string
+	// priorityMap maps the value to the priority.
+	priorityMap map[string]int
+	// elements are the map of elements structured by their priority.
+	elements []map[int]ResponseWatch
+	// numberOfPriorities is the number of priorities in the prioritySets
+	numberOfPriorities int
```
nfuden, line 27: Why do we even care about this? isnt this  just the len of prioritymap?

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
...
+	// nextUniqueElementIndex is the next element index to be inserted.
+	nextUniqueElementIndex int
```
nfuden, line 29: Once again why is this not just a function of length on the list or map?

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
...
+
+func NewPrioritySortedStruct(prioritySets map[int][]string) *PrioritySortedStruct {
+	p := PrioritySortedStruct{
+		prioritySets: prioritySets,
+	}
+	p.Init()
+	return &p
+}
+
+func (p *PrioritySortedStruct) Init() {
```
nfuden, line 46: Why is there a separate init?

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
...
+	p.numberOfPriorities = len(p.prioritySets)
+	// need to ensure that the prioriries are in order and there are no missing or skipped Priorities
+	priorities := make([]int, 0)
+	for priority, _ := range p.prioritySets {
+		priorities = append(priorities, priority)
+	}
+	sort.Ints(priorities)
+	currentP := 0
+	for _, p := range priorities {
```
nfuden, line 55: Why are we enforcing strict numbers on the priorities. If we are doing this we can just have a [][]watch where its len is num prios. Not saying we should but if we were to follow roughly this strategy I dont see why we need to do all this. It feels like its alot more fragile and doesnt really link into the go map/slice semantics

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
...
+		if currentP == p {
+			currentP++
+		} else {
+			panic(fmt.Sprintf("Priorities are not set correct, you are missing priority %d", currentP))
+		}
+	}
+	p.elements = make([]map[int]ResponseWatch, p.numberOfPriorities+1)
+	if p.prioritySets != nil && p.numberOfPriorities > 0 {
```
nfuden, line 63: Why are we doing all this set up ahead of time? Cant we just check if the given prio exists at insert time and add it then?
That being said not against this entirely

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
...
+	for i := 0; i <= p.numberOfPriorities; i++ {
+		m := p.elements[i]
+		for index, v := range m {
+			processFunc(v, PriorityIndex{Priority: i, Index: index})
+		}
+	}
+}
+
+// GetPriorityList returns an ordered list of the elements by priority.
+func (p *PrioritySortedStruct) GetPriorityList() []ResponseWatch {
```
nfuden, line 98: If we are recreating the list each time anyways perhaps we go with the simple section where we just recreate the list from the existing map? It seems like we are adding overhead and complexity here while also diverging more from upstream. 
This function makes me question if I understand what we are trying to do.

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
...
+	nextUniqueElementIndex int
+}
+
+// PriorityIndex is the priority and index used to locate items.
+type PriorityIndex struct {
+	Priority int
+	Index    int
+}
+
+func NewPrioritySortedStruct(prioritySets map[int][]string) *PrioritySortedStruct {
```
nfuden, line 38: Why are we keeping strings here when we could just keep a watch like a map[int][]watch or a [][]Watch

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,165 @@
+package cache
+
+import (
+	"fmt"
+	"sort"
+)
+
+// The default Priority is the last index. The last index is not required in the prioirty list, thus any elements added
+// to the struct will be added to the default priority (last).
+const defaultPrioirty = -1
```
nfuden, line 10: defaultpriority could just be 0 and not a different const?

 see https://go.dev/play/p/k5akHGKRAwf

nfuden (changes requested):

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: 0 is an index in the priority map

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: The strings are the `Cluster` and `Listener`. Look at the Generics in this [PR](https://github.com/solo-io/go-utils/pull/483). The type should be comparable.

jackstine: The items that the collection contains are ResponseWatches.
This is the `elements` property.

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: This is not used at all in the code.
This is if the user of the code wanted to `GetPriorityList` from the struct

Look at the [PR](https://github.com/solo-io/go-utils/pull/483)

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine:
```
	if priority == defaultPrioirty {
		// add to the last index of the watches
		priority = p.numberOfPriorities
	}
```

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: There was [a request](https://github.com/solo-io/go-utils/pull/483#discussion_r892492091) to change the PrioritySet from an `[][]P comparable` to `map[int][]P comparable`

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: I believe this answers your question `Should not have the same index for a deleted element` in the [PR tests](https://github.com/solo-io/go-utils/pull/483/files)

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: to set up the collection

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: Not sure why it matters

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: I think it makes it easy to read

pkg/api/v1/control-plane/cache/prioritylist.go
nfuden: Fair enough.

pkg/api/v1/control-plane/cache/prioritylist.go
jackstine: We can take this out, since so many people have issues with variable here. It can be retrieved with `len(p.prioritySet)`

pkg/api/v1/control-plane/types/types.go
```diff
@@ -0,0 +1,43 @@
+package types
```
nfuden, line 1: I think we should treat this as a breaking change and therefore should have a higher minor version for changelog

jackstine: circular dependency in `pkg/api/v1/control-plane/cache/status.go`

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,178 @@
...
```
nfuden, line 10:
```suggestion
// The default Priority is the last index. The last index is not required in the priority list, thus any elements added
// to the struct will be added to the default priority (last).
const defaultPriority = -1
```

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,178 @@
...
+			elements = append(elements, el)
+		}
+	}
+	return elements
+}
+
+// Add will add the element to the Priority Collection, returns the priority, and element number.
+func (p *PrioritySortedStruct) Add(element ResponseWatch) PriorityIndex {
+	priority := p.getPriorityOfElement(element)
+	if priority == defaultPrioirty {
```
nfuden, line 125:
```suggestion
	if priority == defaultPriority {
```

pkg/api/v1/control-plane/cache/prioritylist.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// getPriorityOfElement returns the priority of element ResponseWatch.
+func (p *PrioritySortedStruct) getPriorityOfElement(element ResponseWatch) int {
+	pv := element.GetPriority()
+	if p, exists := p.priorityMap[pv]; exists {
+		return p
+	} else {
+		// default priority is -1
+		return defaultPrioirty
```
nfuden, line 176:
```suggestion
		return defaultPriority
```

pkg/api/v1/control-plane/cache/simple.go
```diff
@@ -246,15 +268,15 @@ func (cache *snapshotCache) CreateWatch(request Request) (chan Response, func())
...
-				request.TypeUrl, request.ResourceNames, nodeID, request.VersionInfo)
-		}
 		info.mu.Lock()
-		info.watches[watchID] = ResponseWatch{Request: request, Response: value}
+		// check SetSnapshot() for responses on the watches map
+		priority := info.watches.Add(ResponseWatch{Request: request, Response: value})
 		info.mu.Unlock()
-		return value, cache.cancelWatch(nodeID, watchID)
+		if cache.log != nil {
+			cache.log.Infof("open watch Prioiry Index %d and Element Index %d for %s%v from nodeID %q, version %q",
```
nfuden, line 276:
```suggestion
			cache.log.Infof("open watch Priority Index %d and Element Index %d for %s%v from nodeID %q, version %q",
```

nfuden (commented):
a few others with this spelling

nfuden (approved):
Works for me

PR #482, opened 2022-06-16
Configure types to skip schemas

sam-heilbron:
# Blocked By
- [x] https://github.com/solo-io/protoc-gen-openapi/pull/4

# Context
Builds off of https://github.com/solo-io/protoc-gen-openapi/pull/4

### Introduce `MessagesWithEmptySchema`
Expose `MessagesWithEmptySchema` in the codegen.ValidatioSchemaOptions. This is a set of message names, who schema will be ignored, and instead an empty open api spec will be generated. This is useful in the case of recursive protos, or when the validation schema is not needed (ie internal api).

### Remove `MaxDescriptionCharacters`
This field is no longer used. Validation schema descriptions are not useful, and are removed entirely using the `IncludeDescriptionsInSchema` option.

**Technically this is a public facing API so it would constitute a breaking change. However, this library is only used by Gloo Edge, and Edge isn't even using this API (which is why we're removing it), so I decided to just make this a patch.**

# Testing 
`schemagen_test` adds a case where we inject certain messages, and confirm that they have empty schemas.


sam-heilbron:
/kick not sure why build-bot started off in an error state https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=21d29b0b-d376-4921-aab1-eca284cbf2a7

bewebi (approved):
LGTM

PR #491, opened 2022-06-22
Status Marshaller Never Errors

sam-heilbron:
# Description
Status marshaling never errors

In addition to unit tests, this branch has been integrated with a branch of Gloo 1.12 and the intended bugfixing behavior has been verified 

# Context

## Definitions
[Input resource](https://github.com/solo-io/solo-kit/blob/f2a89de3ec965b286a5779f613d2f82c07771438/pkg/api/v1/resources/resource_interface.go#L46) is the base Go type that all top level Gloo CRs implement. 

[Custom input resource](https://github.com/solo-io/solo-kit/blob/f2a89de3ec965b286a5779f613d2f82c07771438/pkg/api/v1/resources/resource_interface.go#L75) is an extension of InputResources, one who explicitly defines how it should be marshaled. **In Gloo Edge, this is useful for resources that need to be compressed before being persisted, like a [Proxy](https://github.com/solo-io/gloo/blob/c89ad1368837712ec072aceaa44a2b6587303079/projects/gloo/pkg/api/v1/proxy_extensions.go#L9)**

[ResourceClient](https://github.com/solo-io/solo-kit/blob/f2a89de3ec965b286a5779f613d2f82c07771438/pkg/api/v1/clients/client_interface.go#L30) is the abstract type used to read and write resources, so some underlying persistence. It abstracts both (A) the type of resources and (B) the type of persistence layer. This resource client provides a generic interface for interacting with resources, and relies on solo-kit to perform the necessary logic to convert the data from the structure it is persisted in, into the structure that Gloo requires.

[ResourceClientFactory](https://github.com/solo-io/solo-kit/blob/f2a89de3ec965b286a5779f613d2f82c07771438/pkg/api/v1/clients/factory/resource_client_factory.go#L42) generates a set of resource clients. This is the central function that can be used to create all resource clients.

[Kube ResourceClient](https://github.com/solo-io/solo-kit/blob/f2a89de3ec965b286a5779f613d2f82c07771438/pkg/api/v1/clients/kube/resource_client.go#L101) is the implementation to read and write resources to Kubernetes.

## Who writes statuses?
Gloo Edge (and any other project that consumes this library) use ResourceClients to update resources with statuses. Specifically, the [resource reporter](https://github.com/solo-io/solo-kit/blob/f2a89de3ec965b286a5779f613d2f82c07771438/pkg/api/v2/reporter/reporter.go#L246) takes a set of reports and writes the data to the status for each relevant resource.

Since Gloo controllers are responsible for maintaining statuses, and not external users, it is fair to expect the persisted status to be properly typed.

## When would status unmarshalling error?
TODO

#  Gloo Edge
This functionality is heavily used by Gloo Edge, and we need to make some corresponding changes in Edge, as well as test them out, before this functionality should merge.

https://github.com/solo-io/gloo/pull/5352 is the PR that introduced this functionality originally in Gloo.


pkg/utils/statusutils/env.go
```diff
@@ -10,6 +10,7 @@ import (
 // to associate a resource status with the appropriate controller statusReporterNamespace
 const PodNamespaceEnvName = "POD_NAMESPACE"
 
+//noinspection GoUnusedExportedFunction
```
jenshu, line 13: is this not used?

jenshu (approved):

PR #497, opened 2022-07-01
fix initial map

EItanya:


sam-heilbron (commented):
The changes themselves look good to me. I think it could be valuable to add a test to validate this behavior (presumably one that would fail without this change), given that its critical to Gloo functioning properly.

nfuden (approved):
https://github.com/solo-io/solo-kit/issues/499 has been made and added to next up to have a living test. 
Because this is a big issue in certain configurations and is shown not to break things we are comfortable moving ahead without the extra test step for the very short term.

PR #502, opened 2022-07-08
[Gloo 1.8] Kube Resource Client Improvements

sam-heilbron:
# Description
Add enhancements to the Kube resource client for the solo-kit version that power Gloo Edge 1.8

# Context

##  Support Expression Selection (https://github.com/solo-io/gloo/issues/6406)
This bug was resolved in later versions of solo-kit, but never made it's way to this branch. [This PR](https://github.com/solo-io/solo-kit/pull/471) added support in later version of solo-kit.

## Allow Unknown Fields (https://github.com/solo-io/solo-kit/issues/505)
When downgrading versions, an older version of Gloo Edge will fail to unmarshal resources with newer APIs. The proposal is to allowUnknownFields. This is safe, because the CRDs still contain a validation schema, so that will reject resources that truly don't fit the spec.

For example,  in a later version of Gloo Edge, we introduce a  field `metadata_static` that was introduced in https://github.com/solo-io/gloo/pull/6037

## Skip Tests on Releases
We historically have used build-bot to execute tests and release this repo. The effect is that tests run on releases and PRs. We've been adding support across all of our libraries to disable tests on releases since they are somewhat flakey, slow down the release process, and don't validate anything that we didn't already validate in a PR.

I opted to backport that functionality in this PR as well.

It works by only running tests when the RELEASE environment variable != true

# Testing
- These changes can be tested internally in this library and I added tests to verify
- This change is pulled into https://github.com/solo-io/gloo/pull/6728 to test in Gloo Edge as well


sam-heilbron:
/kick unclear for reason of timeout https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=44a749ad-58d6-44cb-b14b-eb80e5a85d76

Makefile
```diff
@@ -105,7 +113,9 @@ verify-envoy-protos:
 # as this code is no longer used
 .PHONY: test
 test:
-	PATH=$(DEPSGOBIN):$$PATH ginkgo -r  -v -race -p -tags solokit -compilers=2 -skip multicluster -regexScansFilePath
+ifneq ($(RELEASE), "true")
+	PATH=$(DEPSGOBIN):$$PATH ginkgo -r  -v -race -tags solokit -compilers=2 -skip multicluster -regexScansFilePath -randomizeAllSpecs -randomizeSuites $(TEST_PKG)
+endif
```
jenshu, line 118: won't this cause tests to no longer run in ci?

sam-heilbron: Correct, this will cause tests to run if RELEASE != true.

This way, it will run for PRs, but not for releases. This is what we do on our main branch: https://github.com/solo-io/solo-kit/blob/master/Makefile#L123, as tests can be flaky and they dont' produce a different result than PRs

pkg/api/v1/clients/kube/resource_client_test.go
```diff
@@ -512,11 +520,44 @@ var _ = Describe("Test Kube ResourceClient", func() {
...
+						Namespace: namespace2,
+						Labels: map[string]string{
+							"name":      "res-4",
+							"namespace": namespace2,
+						},
+					},
+				}
+
+				for i, meta := range metadataForMockResources {
+					Expect(util.CreateMockResourceWithMetadata(ctx, clientset, meta, fmt.Sprintf("val-%d", i)))
```
jenshu, line 559:
```suggestion
					Expect(util.CreateMockResourceWithMetadata(ctx, clientset, meta, fmt.Sprintf("val-%d", i))).NotTo(HaveOccurred())
```
should we be checking the error?

jenshu:
LGTM will approve once ci passes

pkg/api/v1/clients/kube/resource_client_test.go
sam-heilbron: you're so right!

pkg/api/v1/clients/kube/resource_client_test.go
```diff
@@ -384,10 +386,15 @@ var _ = Describe("Test Kube ResourceClient", func() {
 				Expect(errors.IsNotExist(err)).To(BeTrue())
 			})
 
-			It("return an error when receiving a malformed resource", func() {
-				_, err := rc.Read(namespace1, malformedResourceName, clients.ReadOpts{})
-				Expect(err).To(HaveOccurred())
-				Expect(errors.IsNotExist(err)).To(BeFalse())
+			It("ignores unknown fields when returning a malformed resource", func() {
```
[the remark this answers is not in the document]
sam-heilbron: yes yes yes! good call

jenshu (approved):

PR #511, opened 2022-07-21
test(emitter): added follow-on test for previous commit

gunnar-solo:
This PR adds a test to simulate a _very specific_/_awkward_ race condition which was addressed in https://github.com/solo-io/solo-kit/pull/497.  The test was added in `test/mocks/v1/non_template_tests_test.go`, with some minor changes to `pkg/code-generator/codegen/templates/resource_client_template.go` in support of it.

The original bug was a result of a race condition between N `.List` and `.Watch` calls against `snapshotEmitters`.  Consider two namespaces: `namespace_1` and `namespace_2`, as well as two `Upstreams` per namespace: `east-1`, `east-2`, `west-1`, `west-2`

```python
# initial population via .List
currentSnapshot.Upstreams = ["east-1", "east-2", "west-1", "west-2"]
upstreamsByNamespace = {}

# namespace_1 .Watch returns
currentSnapshot.Upstreams = ["east-1", "east-2"]
upstreamsByNamespace = {
    "namespace_1": ["east-1", "east-2"]
    "namespace_2": []
}

# namespace_2 .Watch returns
currentSnapshot.Upstreams = ["east-1", "east-2", "west-1", "west-2"]
upstreamsByNamespace = {
    "namespace_1": ["east-1", "east-2"]
    "namespace_2": ["west-1", "west-2"]
}
```

We see in the pre-fix case that there is a period of time between watcher returns in which we _incorrectly_ remove some Upstreams from having been detected.  Though this was theoretically an eventually-consistent bug, it still impacted customers.  Post-fix resembles the following:

```python
# initial population via .List
currentSnapshot.Upstreams = ["east-1", "east-2", "west-1", "west-2"]
upstreamsByNamespace = {
    "namespace_1": ["east-1", "east-2"]
    "namespace_2": ["west-1", "west-2"]
}

# namespace_1 .Watch returns
currentSnapshot.Upstreams = ["east-1", "east-2", "west-1", "west-2"]
upstreamsByNamespace = {
    "namespace_1": ["east-1", "east-2"]
    "namespace_2": ["west-1", "west-2"]
}

# namespace_2 .Watch returns
currentSnapshot.Upstreams = ["east-1", "east-2", "west-1", "west-2"]
upstreamsByNamespace = {
    "namespace_1": ["east-1", "east-2"]
    "namespace_2": ["west-1", "west-2"]
}
```
------------------------

TL;DR:  this is a test which validates https://github.com/solo-io/solo-kit/pull/497.

### DIFF
--- a/pkg/code-generator/codegen/templates/resource_client_template.go
+++ b/pkg/code-generator/codegen/templates/resource_client_template.go
@@ -8,6 +8,9 @@ var ResourceClientTemplate = template.Must(template.New("resource_reconciler").F
 
 import (
 	"context"
+{{- if eq .Name "FakeResource" }}
+	"time"
+{{- end }}
 
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients/factory"
@@ -130,7 +133,7 @@ func (client *{{ lower_camel .Name }}Client) List(namespace string, opts clients
 	if err != nil {
 		return nil, err
 	}
-	return convertTo{{ .Name }}(resourceList), nil
+	return convertTo{{ .Name }}(resourceList, ""), nil
 }
 
 {{ if .ClusterScoped }}
@@ -153,7 +156,11 @@ func (client *{{ lower_camel .Name }}Client) Watch(namespace string, opts client
 			select {
 			case resourceList := <-resourcesChan:
 				select {
-					case {{ lower_camel .PluralName }}Chan <- convertTo{{ .Name }}(resourceList):
+{{- if eq .Name "FakeResource" }}
+					case {{ lower_camel .PluralName }}Chan <- convertTo{{ .Name }}(resourceList, namespace):
+{{- else }}
+					case {{ lower_camel .PluralName }}Chan <- convertTo{{ .Name }}(resourceList, ""):
+{{- end }}
 					case <-opts.Ctx.Done():
 						close({{ lower_camel .PluralName }}Chan)
 						return
@@ -167,7 +174,16 @@ func (client *{{ lower_camel .Name }}Client) Watch(namespace string, opts client
 	return {{ lower_camel .PluralName }}Chan, errs, nil
 }
 
-func convertTo{{ .Name }}(resources resources.ResourceList) {{ .Name }}List {
+func convertTo{{ .Name }}(resources resources.ResourceList, namespace string) {{ .Name }}List {
+{{- if eq .Name "FakeResource" }}
+	if namespace == "slow-watch-namespace" {
+		// This is _only_ utilized by FakeResource, specifically to test out fix for https://github.com/solo-io/gloo/issues/5554.
+		// The general premise is that we need the ability to _conditionally_ have slow watchers to observe
+		// what currentSnapshot.Fakes is over time.  We expect it to _not_ lose data when initial watchers report data.
+		time.Sleep(5 * time.Second)
+	}
+
+{{- end }}
 	var {{ lower_camel .Name }}List {{ .Name }}List
 	for _, resource := range resources {
 		{{ lower_camel .Name }}List = append({{ lower_camel .Name }}List, resource.(*{{ .Name }}))
--- a/pkg/code-generator/codegen/templates/resource_client_template.go
+++ b/pkg/code-generator/codegen/templates/resource_client_template.go
@@ -8,9 +8,6 @@ var ResourceClientTemplate = template.Must(template.New("resource_reconciler").F
 
 import (
 	"context"
-{{- if eq .Name "FakeResource" }}
-	"time"
-{{- end }}
 
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients/factory"
@@ -133,7 +130,7 @@ func (client *{{ lower_camel .Name }}Client) List(namespace string, opts clients
 	if err != nil {
 		return nil, err
 	}
-	return convertTo{{ .Name }}(resourceList, ""), nil
+	return convertTo{{ .Name }}(resourceList), nil
 }
 
 {{ if .ClusterScoped }}
--- a/pkg/code-generator/codegen/templates/resource_client_template.go
+++ b/pkg/code-generator/codegen/templates/resource_client_template.go
@@ -156,11 +156,7 @@ func (client *{{ lower_camel .Name }}Client) Watch(namespace string, opts client
 			select {
 			case resourceList := <-resourcesChan:
 				select {
-{{- if eq .Name "FakeResource" }}
-					case {{ lower_camel .PluralName }}Chan <- convertTo{{ .Name }}(resourceList, namespace):
-{{- else }}
-					case {{ lower_camel .PluralName }}Chan <- convertTo{{ .Name }}(resourceList, ""):
-{{- end }}
+					case {{ lower_camel .PluralName }}Chan <- convertTo{{ .Name }}(resourceList):
 					case <-opts.Ctx.Done():
 						close({{ lower_camel .PluralName }}Chan)
 						return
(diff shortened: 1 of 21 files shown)


sam-heilbron (changes requested):
Could you add to the PR context to outline what the bug was and why this test should fix it?

I created a new branch (https://github.com/solo-io/solo-kit/tree/sam/emitter-map-follow-up-without-fix), where I checked out these changes, and then reverted the commit that fixed the bug. Then I ran:
```
kind create cluster --name solo-kit
TEST_PKG=./test/mocks/v1/  make test
```

I would expect these tests to fail without the fix and pass with them. These tests still passed however, so I think there is more we need to account for

test/mocks/v1/non_template_tests_test.go
```diff
@@ -0,0 +1,194 @@
...
+		mockCustomTypeClient, err = NewMockCustomTypeClient(ctx, mockCustomTypeClientFactory)
+		Expect(err).NotTo(HaveOccurred())
+		// Pod Constructor
+		podClientFactory := &factory.MemoryResourceClientFactory{
+			Cache: memory.NewInMemoryResourceCache(),
+		}
+
+		podClient, err = github_com_solo_io_solo_kit_pkg_api_v1_resources_common_kubernetes.NewPodClient(ctx, podClientFactory)
+		Expect(err).NotTo(HaveOccurred())
+		emitter = NewTestingEmitter(simpleMockResourceClient, mockResourceClient, fakeResourceClient, anotherMockResourceClient, clusterResourceClient, mockCustomTypeClient, podClient)
```
sam-heilbron, line 134: nit: We have somewhat confusing mocks, but there is a v1 package (this one) with many clients, and a v1alpha package with fewer clients. We could use the v1alpha package to avoid needing all the clients that ultimately aren't used by the test

pkg/code-generator/codegen/templates/resource_client_template.go
```diff
@@ -153,7 +156,11 @@ func (client *{{ lower_camel .Name }}Client) Watch(namespace string, opts client
 			select {
 			case resourceList := <-resourcesChan:
 				select {
-					case {{ lower_camel .PluralName }}Chan <- convertTo{{ .Name }}(resourceList):
+{{- if eq .Name "FakeResource" }}
```
sam-heilbron, line 159: I like how you guard this test case so we only generate this for the specific resource that occurs in our test. However, I think it's not ideal that we need to have our template have logic that is specific to a test case.

In the below branch, I outline an alternate proposal, for creating a new client that wraps the existing FakeResourceClient, and injects latency into that client:
https://github.com/solo-io/solo-kit/compare/emitter-map-follow-up...sam/emitter-map-follow-up?expand=1

Let me know what you think!

test/mocks/v1/non_template_tests_test.go
```diff
@@ -0,0 +1,194 @@
...
+			case <-timeout:
+				// clean up created resources
+				err = fakeResourceClient.Delete(fakeResource1a.GetMetadata().Namespace, fakeResource1a.GetMetadata().Name, clients.DeleteOpts{Ctx: ctx})
+				Expect(err).NotTo(HaveOccurred())
+				err = fakeResourceClient.Delete(fakeResource1b.GetMetadata().Namespace, fakeResource1b.GetMetadata().Name, clients.DeleteOpts{Ctx: ctx})
+				Expect(err).NotTo(HaveOccurred())
+				err = kubeutils.DeleteNamespacesInParallelBlocking(ctx, kube, slowWatchNamespace)
+				Expect(err).NotTo(HaveOccurred())
+				break outer
+			case <-ticker:
```
sam-heilbron, line 185: I like the idea of a ticker, though I think we can simplify this to be whenever a new snapshot is pulled off the snapshots channel. Currently the ticker will read every 1/10th of a second, but we won't receive snapshots that quickly so we'll just block on the next line.

test/mocks/v1/non_template_tests_test.go
```diff
@@ -0,0 +1,194 @@
...
+				snap := <-snapshots
+				// 2 Fakes is optimal (everything is found)
+				// 1 Fakes represents the case of a watcher overwriting the initially populated 2 Fakes from .List calls
```
sam-heilbron, line 188: Is it possible for there to be 1 resource initially, when only the first namespace has returned?

test/mocks/v1/non_template_tests_test.go
```diff
@@ -0,0 +1,194 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		timeout := time.After(time.Second * 10)
+		ticker := time.Tick(time.Second / 10)
+	outer:
+		for {
+			select {
+			case <-ctx.Done():
+				break outer
+			case <-timeout:
```
sam-heilbron, line 176: I like this! Another alternative would be to create resources in the before each and clean them up in an after each. Currently we have some duplicate code where resources (namesapces) are cleaned up in the after each. Can we pick one pattern and use it?

sam-heilbron (commented):
Looking good!

test/mocks/v1/non_template_tests_test.go
gunnar-solo: Not with the code as-written.  All of the initial `.List` operations are synchronously run inside the same loop

Screen Shot 2022-08-25 at 3 01 45 PM

pkg/code-generator/codegen/templates/resource_client_template.go
gunnar-solo: pretty thrilled with this suggestion--I (before) didn't have the knowledge on how to make this in-place in the testing file.

The implementation you threw up is quite nice!  I'll steal it

gunnar-solo: oh yeah!
* great point allowing it to be customizable per namespace
* I'm thrilled we no longer need resource-specific templating

pkg/code-generator/codegen/templates/resource_client_template.go
```diff
@@ -8,6 +8,9 @@ var ResourceClientTemplate = template.Must(template.New("resource_reconciler").F
 
 import (
 	"context"
+{{- if eq .Name "FakeResource" }}
```
sam-heilbron, line 11: I think we can get rid of this

pkg/code-generator/codegen/templates/resource_client_template.go
```diff
@@ -130,7 +133,7 @@ func (client *{{ lower_camel .Name }}Client) List(namespace string, opts clients
 	if err != nil {
 		return nil, err
 	}
-	return convertTo{{ .Name }}(resourceList), nil
+	return convertTo{{ .Name }}(resourceList, ""), nil
```
sam-heilbron, line 136: This too

test/mocks/v1/non_template_tests_test.go
```diff
@@ -0,0 +1,241 @@
...
+		err = fakeResourceClient.Delete(fakeResource1b.GetMetadata().Namespace, fakeResource1b.GetMetadata().Name, clients.DeleteOpts{Ctx: ctx})
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("Should not overwrite initial listed resources for non-returned namespace watchers", func() {
+		/*
+			This test relies on some hardcoded logic in `resource_client_template.go`.  In particular:
+			`FakeResourceClient`, when used in conjunction with the `slowWatchNamespace` has an artificial
+			5s delay when getting back the results from .Watch.  The idea behind the test is that
+			we have a normal, unimpeded watch running in parallel with a slow watch to simulate a race condition.
```
sam-heilbron, line 164: I think we can remove this comment, and probably move some of it to the implementation of our slowFakeResourceClient, to include context around what it is, why it exists

test/mocks/v1/non_template_tests_test.go
```diff
@@ -0,0 +1,241 @@
...
+		// create a snapshot observer on "namespace1" and "slowWatchNamespace"
+		snapshots, _, err := emitter.Snapshots([]string{namespace1, slowWatchNamespace}, clients.WatchOpts{
+			Ctx:         ctx,
+			RefreshRate: time.Second,
+		})
+		Expect(err).NotTo(HaveOccurred())
+
+		timeout := time.After(time.Second * 10)
+	outer:
+		for {
```
sam-heilbron, line 179: 👏 this is awesome!

sam-heilbron (commented):
Some final cleanup, but everything looks great!

sam-heilbron (approved):

PR #514, opened 2022-08-15
Ratelimit emitters

kdorosh:
Ratelimit emitters to once per second so Sync loops acting on Snapshot() have time to finish before being cancelled.

### DIFF
--- a/pkg/code-generator/codegen/templates/snapshot_emitter_template.go
+++ b/pkg/code-generator/codegen/templates/snapshot_emitter_template.go
@@ -264,7 +264,10 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
 		initialSnapshot := currentSnapshot.Clone()
 		snapshots <- &initialSnapshot
 
+		needsSync := false
+		// intentionally rate-limited so that our sync loops have time to complete before the next snapshot is sent
 		timer := time.NewTicker(time.Second * 1)
+		defer timer.Stop()
 		previousHash, err := currentSnapshot.Hash(nil)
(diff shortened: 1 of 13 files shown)


pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -275,7 +278,7 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
 			if err != nil {
 				contextutils.LoggerFrom(ctx).Panicw("error while hashing, this should never happen", zap.Error(err))
 			}
-			if previousHash == currentHash {
+			if !needsSync && previousHash == currentHash {
```
nfuden, line 281: When our current and previous hash are the same why would we need to sync? can we add a comment on that

nfuden: Nevermind I see its due to the forceemit change

kdorosh:
still want to explore removing the force emitter altogether and making the 1 second rate limiting timeout configurable

nfuden:
Instead of rate limiting what if we were to keep a queue of needed updates up to some max channel length. Perhaps with rate limiting configurable as a extreme fallback?

kdorosh:
might still be a good idea in the long term. will close for now since i don't suspect it was the root cause of https://github.com/solo-io/gloo/issues/6599

kdorosh:
related fear https://github.com/solo-io/gloo-mesh-enterprise/issues/5107 i still think this is a good idea to merge

but won't fight the team if there's strong objections

PR #515, opened 2022-08-18
Non Watched Namespace Labeled selectors

### LINKED ISSUE #5868: [Migrated] fine-grained selector for watched CRs
Issue has been moved to https://github.com/k8sgateway/k8sgateway/issues/5868

jackstine:
Adding expression label filters to non-watched namespaces.

1. if `watchNamespaces` is empty and `ExpressionSelector` is empty
     - Select all resources in all namespaces
2. If `watchNamespaces` is set and `ExpressionSelector` is empty
    - Select all resources in the `watchNamespaces`
3. If `watchNamespaces` is empty and `ExpressionSelector` is not empty
    - Select all namespaces that fit the labels criteria in the `ExpressionSelector`
        - Watch and List all resources in these namespaces
    - Watch all new namespaces, 
        - If a namespace is created that fits the criteria in the `ExpressionSelector`
            - watch and list all resources in that namespace
4. If `watchNamespaces` is set and `ExpressionSelector` is set
    - Watch all resources in `watchNamespaces`
    - Plus do (3) above
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5868


pkg/multicluster/v1/kubeconfigs_snapshot_emitter.sk.go
```diff
@@ -178,6 +197,165 @@ func (c *kubeconfigsEmitter) Snapshots(watchNamespaces []string, opts clients.Wa
...
+		excludeNamespacesFieldDesciptors = buffer.String()
+
+		namespacesResources, err := k.CoreV1().Namespaces().List(ctx, metav1.ListOptions{FieldSelector: excludeNamespacesFieldDesciptors})
+		if err != nil {
+			return nil, nil, err
+		}
+		allOtherNamespaces := make([]string, len(namespacesResources.Items))
+		for _, ns := range namespacesResources.Items {
+			allOtherNamespaces = append(allOtherNamespaces, ns.Namespace)
+		}
```
kdorosh, line 226: nit; performance optimization. initialize size of list up front with `make` and then assign using index in for loop

pkg/multicluster/v1/kubeconfigs_snapshot_emitter.sk.go
```diff
@@ -178,6 +197,165 @@ func (c *kubeconfigsEmitter) Snapshots(watchNamespaces []string, opts clients.Wa
...
+							_, hit = kubeconfigsByNamespace[namespace]
+							if !hit {
+								newNamespaces = append(newNamespaces, namespace)
+								continue
+							}
+						}
+						if hit {
+							// add a watch for all the new namespaces
+							// REFACTOR
+							for _, namespace := range newNamespaces {
```
kdorosh, line 307: we don't need hit var.. if `newNamespaces` is empty list then we wont loop at all

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -233,9 +250,189 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
...
+
+		var buffer bytes.Buffer
+		for i, ns := range watchNamespaces {
+			buffer.WriteString("metadata.namespace!=")
+			buffer.WriteString(ns)
+			if i < len(watchNamespaces)-1 {
+				buffer.WriteByte(',')
+			}
+		}
+		excludeNamespacesFieldDesciptors = buffer.String()
```
kdorosh, line 270: any limits in field selector length we should be concerned about?

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -171,74 +202,279 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
...
 
-{{- end}}
+{{- range .Resources}}
+{{- if not .ClusterScoped }}
+	{{ lower_camel .PluralName }}ByNamespace := sync.Map{}
+{{- end }}
+{{- end }}
 
-	currentSnapshot := {{ .GoName }}Snapshot{}
+	if ! watchNamespacesIsEmpty || opts.ExpressionSelector == "" {
```
nfuden, line 217: Why are we catching this if we are just ranging over watchNamespaces?
If watchNamespaces is empty then this does nothing as the range has no contents.

jackstine: `watchNamespaces` should be `len(watchedNamespaces) >= 1`

nfuden: gotcha

pkg/api/external/kubernetes/namespace/resource_namespace.go
```diff
@@ -0,0 +1,70 @@
...
+				close(resourceNamespaceChan)
+				return
+			}
+		}
+	}()
+	return resourceNamespaceChan, errorChan, nil
+}
+
+func convertNamespaceListToResourceNamespace(namespaces skkube.KubeNamespaceList) resources.ResourceNamespaceList {
+	l := resources.ResourceNamespaceList{}
```
nfuden, line 65: nit: we know the size so why dont we allocate the slice at that size to begin with?

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
```diff
@@ -171,74 +202,279 @@ func (c *{{ lower_camel .GoName }}Emitter) Snapshots(watchNamespaces []string, o
...
+
+					for _, ns := range resourceNamespaces {
+						// TODO-JAKE are we sure we need this. Looks like there is a cocurrent map read and map write here
+						
+{{- range .Resources }}
+{{- if (not .ClusterScoped) }}
+						// TODO-JAKE we willl only need to do this once, I might be best to keep a set/map of the current
+						// namespaces that are used
+						if _, hit := {{ lower_camel .PluralName }}ByNamespace.Load(ns.Name); !hit {
+							newNamespaces = append(newNamespaces, ns.Name)
```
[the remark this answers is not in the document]
jackstine: This could be a problem if we have a lot of resources, I am thinking of adding a map of visited namespaces, or at least namespaces with watches currently on them.

pkg/api/external/kubernetes/namespace/resource_namespace.go
```diff
@@ -0,0 +1,70 @@
...
+// if we use the list thingy I guess we could try it out, if it does not work, then lets move forward with the base client
+// solution, that uses the kubernetes interface...
+func NewKubeResourceNamespaceLister(kube kubernetes.Interface, cache cache.KubeCoreCache) resources.ResourceNamespaceLister {
+	return &kubeResourceNamespaceLister{
+		namespace: NewNamespaceClient(kube, cache),
+	}
+}
+
+type kubeResourceNamespaceLister struct {
+	namespace skkube.KubeNamespaceClient
```
kdorosh, line 24: nit; can we rename this to `client`?

pkg/api/external/kubernetes/namespace/resource_namespace.go
```diff
@@ -0,0 +1,70 @@
...
+			select {
+			case namespaceList := <-namespaceChan:
+				select {
+				case resourceNamespaceChan <- convertNamespaceListToResourceNamespace(namespaceList):
+				case <-ctx.Done():
+					close(resourceNamespaceChan)
+					return
+				}
+			case <-ctx.Done():
+				close(resourceNamespaceChan)
```
kdorosh, line 56: instead of putting this twice in the function, can we defer this once at the top of the goroutine? that way if we ever add another return case it's harder for the dev to forget to add this again

test/mocks/v1/testing_snapshot_emitter.sk.go
```diff
@@ -227,204 +248,740 @@ func (c *testingEmitter) Snapshots(watchNamespaces []string, opts clients.WatchO
...
-		{
-			simplemocks, err := c.simpleMockResource.List(namespace, clients.ListOpts{Ctx: opts.Ctx, Selector: opts.Selector})
+	simplemocksByNamespace := sync.Map{}
+	mocksByNamespace := sync.Map{}
+	fakesByNamespace := sync.Map{}
+	anothermockresourcesByNamespace := sync.Map{}
+	mctsByNamespace := sync.Map{}
+	podsByNamespace := sync.Map{}
+
+	if !watchNamespacesIsEmpty || opts.ExpressionSelector == "" {
```
kdorosh, line 268: is this right? also rather than invert `watchNamespacesIsEmpty` we could just use `hasWatchedNamespaces`

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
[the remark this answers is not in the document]
jackstine: Where are the 2 logs? I am not sure what you mean by this? we can have a conversation on this.

test/mocks/v1/testing_snapshot_emitter.sk.go
jackstine: Why do you think this is incorrect? This is the backwards compatible.

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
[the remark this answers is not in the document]
nfuden: The 2 logs were just where we hit it in here plus the aggregator at the calling level.

pkg/api/v1/clients/client_interface.go
```diff
@@ -178,3 +178,20 @@ func (o WatchOpts) WithDefaults() WatchOpts {
 	}
 	return o
 }
+
+func TranslateWatchOptsIntoListOpts(wopts WatchOpts) ListOpts {
```
nfuden, line 182: nit: can we move these from this file to one that has a different name. It feels like this is the right package but client_interface as a file is not where I would expect to find it .

Perhaps it could be combined with the current function in options and turned into a utilities file?

jackstine: I am good with that

pkg/api/external/kubernetes/namespace/resource_namespace.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil, nil, err
+	}
+	namespaceChan := namespaceWatcher.ResultChan()
+	resourceNamespaceChan := make(chan resources.ResourceNamespaceList)
+	errorChannel := make(chan error)
+	go func() {
+		defer func() {
+			close(resourceNamespaceChan)
+			close(errorChannel)
+		}()
```
jackstine, line 131: Closing the channels here when an error occurs.

pkg/api/external/kubernetes/namespace/resource_namespace.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+// NewKubeClientResourceNamespaceLister will create a new resource namespace lister that requires the kubernetes client
+// interface.
+func NewKubeClientResourceNamespaceLister(kube kubernetes.Interface) resources.ResourceNamespaceLister {
+	return &kubeClientResourceNamespaceLister{
+		kube: kube,
+	}
+}
```
jackstine, line 34: I am not familiar with the `KubeCoreCache`, which one is the better approach to getting resources?

kdorosh: kube core cache is for shared informers and i think we should generally prefer it for performance

pkg/api/external/kubernetes/namespace/resource_namespace.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+		}
+	}()
+	return resourceNamespaceChan, errorChannel, nil
+}
+
+func (client *kubeClientResourceNamespaceLister) getExcludeFieldSelector(filtered resources.ResourceNamespaceList) string {
+	var buffer bytes.Buffer
+	for i, rns := range filtered {
+		ns := rns.Name
```
kdorosh, line 164: nit; ns is a confusing name

pkg/api/v1/clients/kube/resource_client_factory.go
```diff
@@ -272,9 +311,36 @@ func (f *ResourceClientSharedInformerFactory) Start() {
...
+		informer, err := f.addNewNamespaceToRegistry(ctx, namespace, rc)
+		if err != nil {
+			// TODO-JAKE not sure if we want to panic here or to return an error.  But I feel like panics are ok.
+			contextutils.LoggerFrom(ctx).Panicf("failed to add new namespace to registry: %v", err)
+		}
+		if err := f.kubeController.AddNewInformer(informer); err != nil {
+			contextutils.LoggerFrom(ctx).Panicf("failed to add new informer to kube controller: %v", err)
+		}
+
+	})
```
jackstine, line 337: I think panicking here should not occur.  
The reason is because if the client adds a new namespace, then Gloo fails.

pkg/api/v1/clients/kube/cache/cache.go
```diff
@@ -136,114 +155,144 @@ func NewKubeCoreCacheWithOptions(ctx context.Context, client kubernetes.Interfac
...
-	kubeController := controller.NewController("kube-plugin-controller",
-		controller.NewLockingSyncHandler(k.updatedOccured), informers...,
+	// TODO allows add in the namespaceLister, all the time, or make it an option.
+	// TODO-JAKE the namespace lister is used for something in Gloo, we need to make sure that it is going to
+	// dealt with.  It might have side affects that I am unaware of, so please go and investigate that.
+
+	// TODO-JAKE please look into the Reconciler as well, those might need to updated too. I am not sure though, as they work differently.
+	if len(namesapcesToWatch) == 1 && namesapcesToWatch[0] == metav1.NamespaceAll {
+		k.addNamespaceLister()
+	}
```
jackstine, line 180: going to change this so that it will always create the namespace lister no matter what.
Or we could change this to allow for a flag to set the option.

pkg/api/v1/clients/kube/cache/cache.go
```diff
@@ -136,114 +155,144 @@ func NewKubeCoreCacheWithOptions(ctx context.Context, client kubernetes.Interfac
...
+func (k *kubeCoreCaches) RegisterNewNamespaceCache(namespace string) error {
+	once, _ := k.registerNamespaceLock.LoadOrStore(namespace, &sync.Once{})
+	onceFunc := once.(*sync.Once)
+	// TODO-JAKE should this panic?
+	onceFunc.Do(func() {
+		informers := k.addNewNamespace(namespace)
+		if err := k.kubeController.AddNewListOfInformers(informers); err != nil {
+			contextutils.LoggerFrom(k.ctx).Panicf("failed to add new list of informers to kube controller: %v", err)
+		}
+	})
```
jackstine, line 292: this should not panic

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -152,6 +151,15 @@ func (rc *ResourceClient) Register() error {
...
 }
 
+func (rc *ResourceClient) RegisterNamespace(namespace string) error {
+	err := rc.sharedCache.RegisterNewNamespace(namespace, rc)
+	if err != nil {
+		return err
+	}
+	rc.namespaceWhitelist = append(rc.namespaceWhitelist, namespace)
+	return nil
+}
```
jackstine, line 161: I am thinking we should probably get rid of the white list all together.

pkg/api/v1/clients/kube/cache/cache.go
```diff
@@ -136,114 +155,138 @@ func NewKubeCoreCacheWithOptions(ctx context.Context, client kubernetes.Interfac
...
+	}
+
+	for _, nsToWatch := range namesapcesToWatch {
+		k.addNewNamespace(nsToWatch)
 	}
 
-	kubeController := controller.NewController("kube-plugin-controller",
-		controller.NewLockingSyncHandler(k.updatedOccured), informers...,
+	// since we now allow namespaces to be registered dynamically we will always create the namespace lister
+	k.addNamespaceLister()
```
jackstine, line 174: Always create the namepsace lister. Was thinking that I might get a flag, but I don't think it is needed.

pkg/api/v1/clients/kube/cache/cache.go
```diff
@@ -136,114 +155,138 @@ func NewKubeCoreCacheWithOptions(ctx context.Context, client kubernetes.Interfac
...
-		informers = append(informers, informer)
-		namespaceLister = kubelisters.NewNamespaceLister(informer.GetIndexer())
-	}
-
 	k := &kubeCoreCaches{
 		podListers:       pods,
 		serviceListers:   services,
 		configMapListers: configMaps,
 		secretListers:    secrets,
-		namespaceLister:  namespaceLister,
```
ben-taussig-solo, line 231: Nit: Why do we no longer construct the `namespaceLister` when we initialize `k`?

pkg/api/v1/clients/kube/cache/cache_test.go
```diff
@@ -90,7 +109,6 @@ var _ = Describe("kube core cache tests", func() {
 			})
 
 			It("can list resources for all listers", func() {
-				Expect(cache.NamespaceLister()).To(BeNil())
```
ben-taussig-solo, line 93: Can we update this check in the other direction now that we always initialize a namespace lister? I.E.,
`Expect(cache.NamespaceLister()).NotTo(BeNil())`

test/mocks/v1alpha1/testing_snapshot_emitter.sk.go
```diff
@@ -124,59 +141,268 @@ func (c *testingEmitter) Snapshots(watchNamespaces []string, opts clients.WatchO
...
+							return
+						case mockResourceChan <- mockResourceListWithNamespace{list: mockResourceList, namespace: namespace}:
+						}
+					}
+				}
+			}(namespace)
 		}
+		if len(newlyRegisteredNamespaces) > 0 {
+			contextutils.LoggerFrom(ctx).Infof("registered the new namespace %v", newlyRegisteredNamespaces)
+		}
```
jackstine, line 285: I log the newly registered namespaces with Info, I think this is a good log.  Thing is it prints out for all emitters.

jackstine: Is there a way to get the name of the emitter? Or I could use reflection, to get the emitter type.

pkg/api/v1/clients/kube/cache/cache.go
jackstine: it gets added [at line 179](https://github.com/solo-io/solo-kit/blob/50578c7478c422f920fd589277aae547431a44f8/pkg/api/v1/clients/kube/cache/cache.go#L179). Still in the same function. 
Is this way not preferred?

pkg/api/v1/clients/kube/cache/cache.go
jackstine: I don't think a flag is necessary

pkg/api/v1/clients/kube/cache/cache.go
jackstine: this was changed to a flag, because it would allow the client to choose.  The reason was because Gloo needed the option, based off whether it was using RBAC `namespaced:true`

pkg/api/v1/clients/kube/cache/cache.go
jackstine: Flag is necessary for clients.

pkg/code-generator/codegen/templates/snapshot_emitter_template.go
jackstine: Don't think this is relevant anymore.

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -154,6 +155,17 @@ func (rc *ResourceClient) Register() error {
...
+func (rc *ResourceClient) RegisterNamespace(namespace string) error {
+	err := rc.sharedCache.RegisterNewNamespace(namespace, rc)
+	if err != nil {
+		return err
+	}
+	rc.namespaceLock.Lock()
+	rc.namespaceWhitelist = append(rc.namespaceWhitelist, namespace)
+	rc.namespaceLock.Unlock()
+	return nil
+}
```
jackstine, line 167: should we consider getting rid of this white list?

pkg/api/v1/clients/kube/resource_client_factory.go
```diff
@@ -97,6 +104,9 @@ func NewKubeSharedCacheForConfig(ctx context.Context, cluster string, restConfig
 // and, when started, creates a kubernetes controller that distributes notifications for changes to the watches that
 // have been added to the factory.
 // All direct operations on the ResourceClientSharedInformerFactory are synchronized.
+// Note it might be best to use this as a Singleton instead of a Structure
+// if a Singleton suggestion is proposed, then Register will no long be needed,
+// and we can just use Register New Namespace.  Also this would clear up some mutexes, that are being used.
```
jackstine, line 109: Do we want to consider this idea, or at least consider registering and starting the resource client factory on `NewFunction()`?

test/mocks/v1alpha1/testing_snapshot_emitter.sk.go
```diff
@@ -124,59 +141,269 @@ func (c *testingEmitter) Snapshots(watchNamespaces []string, opts clients.WatchO
...
+						name := key.(string)
+						if _, hit := mapOfResourceNamespaces[name]; !hit {
+							missingNamespaces = append(missingNamespaces, name)
+						}
+						return true
+					})
+
+					for _, ns := range missingNamespaces {
+						mockResourceChan <- mockResourceListWithNamespace{list: MockResourceList{}, namespace: ns}
+					}
```
jackstine, line 348: If a namespace is deleted, this causes the namespace to be empty.  The entry in the map is never deleted.  I think this is fine. 

Deleting the entry would require a `RemoveRegisteredNamespace()` for the clients, and a deletion from the Kube-Controller. Since this was never handled originally, I think this is fine.  Since clients make upgrades as well, I think this is fine as too.

pkg/api/v1/clients/kube/cache/cache.go
ben-taussig-solo: Eh doesn't really matter either way

jackstine:
adding this closed PR https://github.com/solo-io/gloo/pull/7097

PR #517, opened 2022-09-09
Resource client patch for apply status

kdorosh:
lowers the number of api calls made to k8s apiserver; helps with status scalability for gloo as a client not getting rate limited talking to the k8s apiserver.

~only marked with WIP label for now to prevent merges until https://github.com/solo-io/gloo/pull/7133 passes ci; in the meantime this is ready for review~ gloo ci is passing


kdorosh:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=7797520a-c807-46e4-b2cf-1fe285a8c709

pkg/api/v2/reporter/reporter.go
```diff
@@ -272,74 +270,31 @@ func (r *reporter) WriteReports(ctx context.Context, resourceErrs ResourceReport
...
+			return r.attemptUpdateStatus(ctx, client, resourceToWrite, status)
 		})
 
+		if errors.IsNotExist(writeErr) {
+			logger.Debugf("did not write report for %v : %v because resource was not found", resourceToWrite.GetMetadata().Ref(), status)
+			delete(resourceErrs, resource)
+			continue
+		}
+
 		if writeErr != nil {
```
jackstine, line 283: What is different about the 2 if statements here?  I am unfamiliar with `errors.IsNotExist(writeErr)`.

pkg/api/shared/status.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	ns := ""
+	for loopNs := range inputResource.GetNamespacedStatuses().GetStatuses() {
+		ns = loopNs
+	}
+	status := inputResource.GetNamespacedStatuses().GetStatuses()[ns]
+
+	buf := &bytes.Buffer{}
+	var marshaller jsonpb.Marshaler
+	marshaller.EnumsAsInts = false  // prefer jsonpb over encoding/json marshaller since it renders enum as string not int (i.e., state is human-readable)
```
elcasteel, line 60: Currently we write out ints for the states. If we want to switch to writing strings we should  call it out in the changelog. I assume that this change won't affect any tooling around statuses because the unmarshaller will handle it the same regardless.

kdorosh: happy to call that out in the changelog. yeah we currently use the jsonpb to unmarshal so it can handle either format, so im just opting for better readability

pkg/api/v2/reporter/reporter.go
kdorosh: that error is for the scenario when we try to patch a resource that no longer exists (e.g. a user deleted it). rather than report an error in the logs, we just remove it from our in memory report and swallow the non-error. for any other error type we will report the error

pkg/api/v1/clients/kubesecret/resource_client.go
```diff
@@ -193,6 +195,37 @@ func (rc *ResourceClient) Write(resource resources.Resource, opts clients.WriteO
...
+	data, err := shared.GetJsonPatchData(inputResource)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting status json patch data")
+	}
+	secret, err := rc.Kube.CoreV1().Secrets(namespace).Patch(opts.Ctx, name, types.JSONPatchType, data, metav1.PatchOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, errors.NewNotExistErr(namespace, name, err)
+		}
+		return nil, errors.Wrapf(err, "patching secret from kubernetes")
```
elcasteel, line 215: Clarify in this message that we're patching the status on the secret. Users tend to zero in on log messages that have things like "secret" in them so it's good to be as clear as possible.

elcasteel: Might as well include "status" in the other error messages well though.

elcasteel (approved):

PR #519, opened 2022-09-20
custom hash for custom resources using spec

yuval-k:


test/mocks/v1/testing_snapshot_emitter_test.go
```diff
@@ -1,5 +1,3 @@
-// Code generated by solo-kit. DO NOT EDIT.
-
```
kdorosh, line 2: why did this get removed? i think we would want this to stay yes?

yuval-k: if it were generated CI would have failed...

PR #520, opened 2022-09-21
Faster memory client

elcasteel:
Add an implementation of the memory resource client that does not clone resources as they are written and read. In gloo edge, we use the memory client to hold proxies that are only generated by the gloo pod. Because proxies can be quite large, the clones take up significant space and GC resources and we can make decisions about when clones are necessary outside of the client.


EItanya (commented):
Since the behavior of this type is exactly the same without the clone, should the other client extend this one so we ensure the other behavior remains the same?

elcasteel:
I think we need locks around more of the methods in this client - we should not be able to call `Write` and `List` at the same time. 
Also, I was planning to let the caller clone resources that it intends to modify after reading but given that we've recently learned that clones can modify the original, I'm going to add methods to return a cloned resource.

PR #524, opened 2022-09-28
Truncate status in reporter

kdorosh:
Truncate size of statuses written to k8s / etcd to protect against large keys stored in k8s backend.

### DIFF
--- a/pkg/api/v2/reporter/reporter.go
+++ b/pkg/api/v2/reporter/reporter.go
@@ -376,6 +376,11 @@ func trimStatusForMaxSize(status *core.Status, bytesPerKey, maxKeys int) *core.S
 	}
 
 	for key, childStatus := range status.SubresourceStatuses {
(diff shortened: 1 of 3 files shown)


elcasteel:
We should document that this overrides a value set in `PROXY_STATUS_MAX_SIZE_BYTES` We should also deprecate that in future releases.

pkg/api/v2/reporter/reporter.go
```diff
@@ -339,3 +347,41 @@ func (r *reporter) StatusFromReport(report Report, subresourceStatuses map[strin
 		Messages:            messages,
 	}
 }
+
+func trimStatus(status *core.Status) *core.Status {
```
elcasteel, line 351: what's the purpose of this layer of indirection?

kdorosh: just to allow us a place to call with the default values for max bytes / keys. then when can call the other more specific version while recursing to ensure we have a finite size for the reason before writing to etcd

elcasteel (approved):

PR #525, opened 2022-09-29
Omit status if too large

kdorosh:
Skip the status patch if it's too large to protect etcd and k8s communication.

Related to https://github.com/solo-io/solo-kit/pull/524 which attempts to truncate status so we can still write. If that fails (e.g. opaque struct field is very large) we should still short-circuit and skip writing status to protect etcd

### DIFF
--- a/pkg/api/shared/status.go
+++ b/pkg/api/shared/status.go
@@ -2,12 +2,22 @@ package shared
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 
 	"github.com/golang/protobuf/jsonpb"
+	"github.com/solo-io/go-utils/contextutils"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
 	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
 	"github.com/solo-io/solo-kit/pkg/errors"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	// the patch metadata is capped at 2kb of data, for more see https://github.com/solo-io/solo-kit/issues/523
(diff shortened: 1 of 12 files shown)


pkg/api/shared/status.go
```diff
@@ -67,5 +77,15 @@ func GetJsonPatchData(inputResource resources.InputResource) ([]byte, error) {
 	bytes := buf.Bytes()
 	patch := fmt.Sprintf(`[{"op": "replace", "path": "/status/statuses/%s", "value": %s}]`, ns, string(bytes)) // only replace our status so other reporters are not affected (e.g. blue-green of gloo)
 	data := []byte(patch)
+
+	if len(data) > MaxStatusBytes {
+		if contextutils.GetLogLevel() == zapcore.DebugLevel {
```
nfuden, line 82: Should we put an informative statement that the status was too large as the status?

pkg/api/v1/clients/kube/resource_client_test.go
```diff
@@ -315,27 +319,32 @@ var _ = Describe("Test Kube ResourceClient", func() {
 
 			It("call read", func() {
 				_, err := rc.Read(namespace2, "test", clients.ReadOpts{})
-				Expect(err).To(HaveOccurred())
+				Expect(err).To(MatchError(ContainSubstring("this client was not configured to access resources in the")))
```
nfuden, line 322: nit: Can we extract this out to a var?

elcasteel (approved):

pkg/api/shared/status.go
kdorosh: not sure what you wanted here?

PR #527, opened 2022-10-04
Validation Webhook Changes

jackstine:
Resource now interface with more options under a Hashable interface. Snapshots now provide methods to add and get lists based off the type of resource.

### DIFF
--- a/pkg/api/v1/resources/resource_interface.go
+++ b/pkg/api/v1/resources/resource_interface.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"fmt"
+	"hash"
 	"reflect"
 	"sort"
 
@@ -78,6 +79,18 @@ type CustomInputResource interface {
 	UnmarshalStatus(status v1.Status, defaultUnmarshaler StatusUnmarshaler)
 	MarshalSpec() (v1.Spec, error)
 	MarshalStatus() (v1.Status, error)
+}
+
+// Hashable is an interface used for hashing the struture.
+type Hashable interface {
+	Hash(hasher hash.Hash64) (uint64, error)
+	MustHash() uint64
+}
+
+// HashableInputResources are InputResources that can be hashed
+type HashableInputResource interface {
+	InputResource
+	Hashable
 }
 
 type ResourceList []Resource
--- a/pkg/code-generator/codegen/templates/resource_template.go
+++ b/pkg/code-generator/codegen/templates/resource_template.go
@@ -38,12 +38,19 @@ var (
 {{- if $.IsCustom }}
 	_ resources.CustomInputResource = new({{ .Name }})
 {{- else }}
-	_ resources.InputResource = new({{ .Name }})
+	_ resources.HashableInputResource = new({{ .Name }})
 {{- end }}
 {{- else }}
 	_ resources.Resource = new({{ .Name }})
 {{- end }}
 )
+{{- if $.HasStatus }}
+{{- if not $.IsCustom }}
+func New{{ .Name }}HashableInputResource() resources.HashableInputResource {
+	return new({{ .Name }})
+}
+{{- end }}
+{{- end }}
 
 func New{{ .Name }}(namespace, name string) *{{ .Name }} {
 	{{ lowercase .Name }} := &{{ .Name }}{}
--- a/pkg/code-generator/codegen/templates/snapshot_template.go
+++ b/pkg/code-generator/codegen/templates/snapshot_template.go
@@ -7,6 +7,16 @@ import (
 var ResourceGroupSnapshotTemplate = template.Must(template.New("resource_group_snapshot").Funcs(Funcs).Parse(
 	`package {{ .Project.ProjectConfig.Version }}
 
+{{/* creating a variable that lets us understand how many resources are hashable input resources. */}}
+{{- $num_of_custom_resources := 0 }}
+{{- range .Resources }}
+{{- if .HasStatus }}
+{{- if not .IsCustom }}
+	{{ $num_of_custom_resources = inc $num_of_custom_resources }}
+{{- end }}
+{{- end }}
+{{- end }}
+
 import (
 	"encoding/binary"
 	"fmt"
@@ -15,6 +25,7 @@ import (
 	"log"
 
 	{{ .Imports }}
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
 	"github.com/rotisserie/eris"
 	"github.com/solo-io/go-utils/hashutils"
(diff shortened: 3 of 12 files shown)


pkg/api/v1/resources/resource_interface.go
```diff
@@ -80,6 +81,18 @@ type CustomInputResource interface {
...
 
+// Hashable is an interface used for hashing the struture.
+type Hashable interface {
+	Hash(hasher hash.Hash64) (uint64, error)
+	MustHash() uint64
+}
+
+// HashableInputResources are InputResources that can be hashed
+type HashableInputResource interface {
+	InputResource
```
sam-heilbron, line 92: 👏

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -38,12 +38,17 @@ var (
 {{- if $.IsCustom }}
 	_ resources.CustomInputResource = new({{ .Name }})
```
sam-heilbron, line 39: CustomInputResources extend Inputresources (things with statuses) by allowing the definition of custom marshaling behavior. if we change this template to create a hashableInputResource, i think we'll need to update the CustomInputResource as well

pkg/code-generator/codegen/templates/resource_template.go
```diff
@@ -38,12 +38,17 @@ var (
...
 {{- else }}
-	_ resources.InputResource = new({{ .Name }})
+	_ resources.HashableInputResource = new({{ .Name }})
 {{- end }}
 {{- else }}
 	_ resources.Resource = new({{ .Name }})
 {{- end }}
 )
+{{- if not $.IsCustom }}
```
sam-heilbron, line 47: I'm curious about this guard, why don't we want to support custom input resources as well?

changelog/v0.31.0/hashable-input-resources.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: BREAKING_CHANGE
```
sam-heilbron, line 2: If this is breaking, how should consumers of this library upgrade? Can we call out in the PR how a developer in gloo for example would bring in this change?

pkg/code-generator/codegen/templates/snapshot_template.go
```diff
@@ -80,6 +93,103 @@ func (s {{ .GoName }}Snapshot) HashFields() []zap.Field {
...
+		s.{{ upper_camel .PluralName }}.Sort()
+		return nil
+{{- end }}
+{{- end }}
+	default:
+		return eris.New("did not add the input resource type because it does not exist")
+	}
+}
+
+func (s *{{.GoName}}Snapshot) ReplaceResource(i int, resource resources.Resource) error {
```
sam-heilbron, line 126: I'm a little confused by this API. We are exposing the ability to replace a resource in a snapshot at a particualr index. How does the consumer of this library determine what i should be? 

Can you help me understand the use case where this would be important? I fear that our snapshots originally had a contract that they wouldn't be modified, so I want to make sure that if we're introducing support for that we consider all the eeffects

test/mocks/v2alpha1/testing_snapshot.sk.go
```diff
@@ -86,6 +88,82 @@ func (s TestingSnapshot) HashFields() []zap.Field {
 	return append(fields, zap.Uint64("snapshotHash", snapshotHash))
 }
 
+func (s *TestingSnapshot) GetResourcesList(resource resources.Resource) (resources.ResourceList, error) {
```
sam-heilbron, line 91: How does this API differ from using s.Mock.AsResources directly?

pkg/code-generator/codegen/templates/resource_template.go
jackstine: based off your other comment, I had Initially placed this as `HashableInputResource`.

test/mocks/v2alpha1/testing_snapshot.sk.go
jackstine: When you don't know the type of `resource`.

pkg/code-generator/codegen/templates/snapshot_template.go
jackstine: So in gloo we make a clone of the snapshot, and do the same thing. We will replace the resource in the list or add to the list based off if the resource we are validating already exists in the snapshot.
This is done in the [apply function](https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/validation/validator.go#L470)

pkg/code-generator/codegen/templates/resource_template.go
jackstine: I think it is reasonable to get rid of the hashable input resource, or we could make `InputResource` interface with `HashableResource` instead of `Resource`.

pkg/code-generator/codegen/templates/snapshot_template.go
```diff
@@ -80,6 +83,85 @@ func (s {{ .GoName }}Snapshot) HashFields() []zap.Field {
...
+			if refKey != res.GetMetadata().Ref().Key() {
+				newList = append(newList, res)
+			}
+		}
+		s.{{ upper_camel .PluralName }} = newList
+		s.{{ upper_camel .PluralName }}.Sort()
+		return nil	
+{{- end }}
+	default:
+		return eris.Errorf("did not remove the reousource because its type does not exist [%T]", resource)
```
kdorosh, line 113: nit; typo
```suggestion
		return eris.Errorf("did not remove the resource because its type does not exist [%T]", resource)
```

pkg/code-generator/codegen/templates/snapshot_template.go
```diff
@@ -80,6 +83,85 @@ func (s {{ .GoName }}Snapshot) HashFields() []zap.Field {
...
+	}
+}
+
+func (s *{{ .GoName }}Snapshot) AddOrReplaceToResourceList(resource resources.Resource) error {
```
kdorosh, line 117: nit; i think upsert is slightly shorter / more readable than `AddOrReplace`, but if this is a painful refactor or others disagree i don't feel super strongly about this

pkg/code-generator/codegen/templates/snapshot_template.go
```diff
@@ -80,6 +83,85 @@ func (s {{ .GoName }}Snapshot) HashFields() []zap.Field {
...
+		}
+		s.{{ upper_camel .PluralName }}.Sort()
+		return nil	
+{{- end }}
+	default:
+		return eris.Errorf("did not add/replace the resource type because it does not exist %T", resource)
+	}
+}
+
+func (s *{{ .GoName }}Snapshot) AddToResourceList(resource resources.Resource) error {
```
kdorosh, line 140: why would we ever want this as opposed to the upsert? this would allow us to add duplicate resources to a snapshot which seems always wrong

pkg/code-generator/codegen/templates/snapshot_template.go
```diff
@@ -80,6 +83,85 @@ func (s {{ .GoName }}Snapshot) HashFields() []zap.Field {
...
+		s.{{ upper_camel .PluralName }} = append(s.{{ upper_camel .PluralName }}, typed)
+		s.{{ upper_camel .PluralName }}.Sort()
+		return nil
+{{- end }}
+	default:
+		return eris.Errorf("did not add the resource type because it does not exist %T", resource)
+	}
+}
+
+func (s *{{.GoName}}Snapshot) ReplaceResource(i int, resource resources.Resource) error {
```
kdorosh, line 153: is this actually used? it seems like we could just do this with assignments on the generated struct

kdorosh (commented):
lgtm overall, just nits!

pkg/code-generator/codegen/templates/snapshot_template.go
jackstine: agreed upsert is better

pkg/code-generator/codegen/templates/snapshot_template.go
jackstine: no this is not used, thought it would be nice to add it anyways.  I can go ahead and get rid of it.

pkg/code-generator/codegen/templates/snapshot_template.go
jackstine: yeah sounds good

changelog/v0.31.0/hashable-input-resources.yaml
jackstine: no longer a breaking change

nfuden (approved):
lgtm

PR #535, opened 2023-01-25
don't error on multiple statuses

jenshu:
Fix an error that would occur if a resource contained namespaced statuses for multiple namespaces.
`GetJsonPatchData` now checks the POD_NAMESPACE environment variable to determine which namespace to modify, instead of erroring.


jenshu:
/kick data race? https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=57ae6d95-4aa7-4548-b12c-d0e327da7062

cloudbuild.yaml
```diff
@@ -7,12 +7,10 @@ steps:
   dir: &dir '/workspace/solo-kit'
 
 # prepare-workspace to set up the project so it can be built and tested
-- name: 'gcr.io/$PROJECT_ID/prepare-go-workspace:0.2.2'
+- name: 'gcr.io/$PROJECT_ID/prepare-go-workspace:0.4.28'
```
jenshu, line 10: cloud-builders bumps were needed to fix the [gke auth plugin](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke) errors in ci

pkg/api/v1/clients/kube/resource_client_factory_test.go
```diff
@@ -166,7 +166,7 @@ var _ = Describe("Test ResourceClientSharedInformerFactory", func() {
 
 			It("correctly handles multiple events", func() {
 
-				var watchResults []string
+				watchResults := NewWatchResults()
```
jenshu, line 169: the changes in this file were to fix a data race in ci

nfuden (approved):
https://github.com/solo-io/gloo/pull/7742 has passed with this referenced
General approach is reasonable and dont see anything objectionable

elcasteel (approved):

PR #536, opened 2023-02-01
Go 1.18

sam-heilbron:
Upgrade go to 1.18 so that we can upgrade Ginkgo from v1 to v2.

Based on https://github.com/solo-io/gloo/pull/6330/

## Breaking Change
Techincally, this isn't a breaking change, but I figured when bumping go versions in a library we should change Minor versions just to be safe.

I intend to to the ginkgo v2 upgrade (which depends on this) as a fast follow. That is a breaking change, and I figured both could go out together.

### DIFF
--- a/cloudbuild-cache.yaml
+++ b/cloudbuild-cache.yaml
@@ -10,17 +10,17 @@ steps:
     path: /root/.ssh
   id: 'untar-mod-cache'
 
-- name: 'golang:1.16.3'
-  args: ['go', 'mod', 'download']
+- name: 'golang:1.18.1'
+  args: ['go', 'mod', 'download', 'all']
   volumes: *vol
   id: 'download'
 
-- name: 'golang:1.16.3'
+- name: 'golang:1.18.1'
   args: ['go', 'mod', 'tidy']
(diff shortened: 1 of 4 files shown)


sam-heilbron:
/skip-changelog

Techincally, this isn't a breaking change, but I figured when bumping go versions in a library we should change Minor versions just to be safe.

I intend to to the ginkgo v2 upgrade (which depends on this) as a fast follow. That is a breaking change, and I figured both could go out together.

cloudbuild-cache.yaml
```diff
@@ -10,17 +10,17 @@ steps:
     path: /root/.ssh
   id: 'untar-mod-cache'
 
-- name: 'golang:1.16.3'
-  args: ['go', 'mod', 'download']
+- name: 'golang:1.18.1'
```
jenshu, line 13: do we want to do 1.18.2 to match what we use in other repos?

sam-heilbron: good call! updated

jenshu (approved):

nfuden (approved):

PR #537, opened 2023-02-02
Ginkgo v2

sam-heilbron:
# Description
Upgrade ginkgo library to use v2

This PR cannot be merged until the following code has been merged and released:
- [x] https://github.com/solo-io/go-utils/pull/506
- [x] https://github.com/solo-io/k8s-utils/pull/32

# Context
I try to order the sections in the order that I resolved them, as they will hopefully work as a template for updating other repos to use Ginkgo v2.

I followed https://onsi.github.io/ginkgo/MIGRATING_TO_V2

## Update Dependency
In order to successfully update the ginkgo v2 library, we needed to update a number of other libraries. I decided to try to update many internal libraries to have them match our Gloo Edge dependencies. The notable upgrades are called out in the changelog

## Update CLI
I updated our Makefile to install v2 locally, and set the appropriate flags when running locally.

Now, all necessary flags are provided by default, but a user can define `GINKGO_USER_FLAGS` to inject others at runtime. 

Also expose a `test-with-coverage` option.

## Update Tests to Import Ginkgo v2
A simple find/replace

## Remove Custom Reporters
The reporting infrastructure (https://onsi.github.io/ginkgo/MIGRATING_TO_V2#improved-reporting-infrastructure) improved. It also no longer supports custom reporters, so I removed any instances of these.

## Remove Multicluster
The multicluster package was not used by solo-kit. Solo-Kit is only consumed by Gloo Edge which doesn't support multicluster features. The Gloo Federation that ships with Gloo Edge Enterprise relies on skv2 to use multicluster functionality. I reached out in Slack to confirm this with other devs.

## Remove Locking Test Utility
While deleting the multicluster packages, I found many suites that used a locking mechanism for a cluster which was unnecessary. In most situations the locking occurred within a BeforeSuite, which is why BeforeSynchronizedSuite is needed. I deleted this and the tests still pass.

## Update Tests to Not Depend on k8s-utils
When using ginkgov2, you cannot mix v1 and v2 libraries (https://onsi.github.io/ginkgo/MIGRATING_TO_V2#can-i-mix-ginkgo-v1-and-ginkgo-v2). Therefore, for any tests which depended on k8s-utils, which still uses ginkgo v1, we needed to find an alternative. The options were:
1. Upgrade k8s-utils to use ginkgov2
2. Update the test to not use that code
3. Copy the utility code into the repo, and reference that in the test

I originally thought option `#3` was the most direct so I went with that option. However, upgrading k8s-utils was trivial so I ended up using option `#1`

## Race Detection
I added the ginkgo `-race` flag to be run, which identifies when we encounter races in our tests. 

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=a52e1a9f-db0e-472e-afc6-c21f5e0ea0e6 encountered a race in our tests. My steps to fix it were:

1. `TEST_PKG=./pkg/api/v1/clients/kube GINKGO_USER_FLAGS=--until-it-fails  RUN_KUBE_TESTS=1  make test` to reproduce the race locally
2. Read the code and the stack track to try to parse the fix
3. Push up the fix
4. Re-run the command from earlier and verify a race does not occur

## Reporter Test
There was a flake in the reporter test. This occurred because we randomize the tests and it illustrated that we weren't correctly setting up the reporter tests.

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=55b3fba0-9a17-46e0-864c-dd0c26f5b896

You can see the solution, thanks to @jenshu was to initialize our `disableTruncateStatus` before all tests, and let each one override it.

# Testing
The tests which encountered the most failures (flakes, races) were the resource client and factory tests. Therefore, to be more certain that they were in a good state, I would run 
```
TEST_PKG=./pkg/api/v1/clients/... GINKGO_USER_FLAGS=--until-it-fails  RUN_KUBE_TESTS=1  make test
```

This would run all the client tests repeatedly and once I could consistently see this pass 10 times in a row, I felt comfortable about the state.


sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=cb9e5199-4350-470a-9cce-5bb526be7441

```
Step #6 - "test": [38;5;10m•[0m[38;5;10m•[0m[38;5;10m•[0m[38;5;10m•[0m[38;5;10m•[0m[38;5;10m•[0m
Step #6 - "test": [38;5;243m------------------------------[0m
Step #6 - "test": [38;5;9m• [FAILED] [1.113 seconds][0m
Step #6 - "test": [0mTest ResourceClientSharedInformerFactory [38;5;243mcreating watches [0mcontext cancellation [38;5;9m[1m[It] watches stop receiving events after the factory's context is cancelled[0m
Step #6 - "test": [38;5;243m/workspace/solo-kit/pkg/api/v1/clients/kube/resource_client_factory_test.go:259[0m
Step #6 - "test": 
Step #6 - "test":   [38;5;9m[FAILED] Timed out after 1.001s.
Step #6 - "test":   Expected
Step #6 - "test":       <int>: 82
Step #6 - "test":   to equal
Step #6 - "test":       <int>: 83[0m
Step #6 - "test":   [38;5;9mIn [1m[It][0m[38;5;9m at: [1m/workspace/solo-kit/pkg/api/v1/clients/kube/resource_client_factory_test.go:295[0m [38;5;243m@ 02/03/23 01:44:29.9[0m
Step #6 - "test": [38;5;243m------------------------------[0m
```

I ran the test locally and it passed.

test/tests/generic/test_crud_client.go
```diff
@@ -16,6 +16,7 @@ import (
 )
 
 // Call within "It"
+// TestCrudClient
```
nfuden, line 19: ?

test/mocks/v2alpha1/kube/client/clientset/versioned/fake/register.go
```diff
@@ -29,7 +29,7 @@ import (
 
 var scheme = runtime.NewScheme()
 var codecs = serializer.NewCodecFactory(scheme)
-var parameterCodec = runtime.NewParameterCodec(scheme)
+
```
nfuden, line 32: why is this being removed? is this file even used?

pkg/code-generator/codegen/templates/resource_client_test_template.go
```diff
@@ -4,15 +4,16 @@ import (
 	"text/template"
 )
 
-var ResourceClientTestTemplate = template.Must(template.New("resource_client_test").Funcs(Funcs).Parse(`// +build solokit
+var ResourceClientTestTemplate = template.Must(template.New("resource_client_test").Funcs(Funcs).Parse(`
+// go:build solokit
```
nfuden, line 8:
```suggestion
//go:build solokit
```

pkg/code-generator/codegen/templates/event_loop_test_template.go
```diff
@@ -4,7 +4,8 @@ import (
 	"text/template"
 )
 
-var ResourceGroupEventLoopTestTemplate = template.Must(template.New("resource_group_event_loop_test").Funcs(Funcs).Parse(`// +build solokit
+var ResourceGroupEventLoopTestTemplate = template.Must(template.New("resource_group_event_loop_test").Funcs(Funcs).Parse(`
+// go:build solokit
```
nfuden, line 8:
```suggestion
//go:build solokit
```

test/tests/generic/test_crud_client.go
sam-heilbron: ah started to update a function comment and then got distracted.Thanks for the callout, ILl update

test/mocks/v2alpha1/kube/client/clientset/versioned/fake/register.go
sam-heilbron: This is auto-generated by our clientset codegen: https://github.com/solo-io/solo-kit/blob/master/Makefile#L84. I would have to look further into how it is actually used in the solo-kit repo. This is updated because I updated the kube libraries we're using here since they were well behind our other projects. 

I think if we want to update/remove this clientset code, it should be done separately. Are you ok with the change as is, due to the library update?

bewebi (commented):
It's a lot to go over exhaustively, so I'm taking it on faith to some extent that the changes that occur in many places are consistent

Have a couple other minor questions as well

pkg/api/external/kubernetes/customresourcedefinition/customresourcedefinition_base_client_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 var _ = Describe("DeploymentBaseClient", func() {
 	if os.Getenv("RUN_KUBE_TESTS") != "1" {
-		log.Printf("This test creates kubernetes resources and is disabled by default. To enable, set RUN_KUBE_TESTS=1 in your env.")
+		fmt.Print("This test creates kubernetes resources and is disabled by default. To enable, set RUN_KUBE_TESTS=1 in your env.")
```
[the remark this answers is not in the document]
sam-heilbron: I originally did this to avoid referencing k8s-utils in our tests. This is because the k8s library still was using ginkgov1 wand that causes tests to panic (called out in last section of PR). I can actually revert this, as I ended up updating k8s-utils as part of this work

Makefile
```diff
@@ -102,31 +127,49 @@ generated-code: $(OUTPUT_DIR)/.generated-code update-licenses
...
 
 #----------------------------------------------------------------------------------
-# Unit Tests
+# Tests
 #----------------------------------------------------------------------------------
 
-# '-skip=multicluster', '-regexScansFilePath' skips any filepath which includes multicluster, which is useful
-# as this code is no longer used
+GINKGO_VERSION ?= 2.5.0 # match our go.mod
+GINKGO_ENV := GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore ACK_GINKGO_DEPRECATIONS=$(GINKGO_VERSION)
```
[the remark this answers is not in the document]
sam-heilbron: Good call

pkg/api/v1/clients/file/file_suite_test.go
```diff
@@ -3,11 +3,10 @@ package file_test
 import (
 	"testing"
 
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
-// TODO: fix tests
```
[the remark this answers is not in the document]
sam-heilbron: I believe so, this comment seemed out of date as the tests pass

pkg/utils/kubeutils/metadata.go
```diff
@@ -39,7 +39,6 @@ func ToKubeMetaMaintainNamespace(meta *core.Metadata) metav1.ObjectMeta {
 	return metav1.ObjectMeta{
 		Name:            meta.Name,
 		Namespace:       meta.Namespace,
-		ClusterName:     meta.Cluster,
```
bewebi, line 42: Why do we do this?

Makefile
```diff
@@ -1,14 +1,36 @@
...
+# To update a target to be self-documenting (and appear with the `help` command),
+# place a comment after the target that is prefixed by `##`. For example:
+#	custom-target: ## comment that will appear in the documentation when running `make help`
+#
+# **NOTE TO DEVELOPERS**
+# As you encounter make targets that are frequently used, please make them self-documenting
+.PHONY: help
+help: FIRST_COLUMN_WIDTH=35
+help: ## Output the self-documenting make targets
+	@grep -hE '^[%a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-$(FIRST_COLUMN_WIDTH)s\033[0m %s\n", $$1, $$2}'
```
gunnar-solo, line 18: woo!  color-coding!

Screen Shot 2023-02-07 at 9 10 27 AM

Makefile
```diff
@@ -102,31 +127,50 @@ generated-code: $(OUTPUT_DIR)/.generated-code update-licenses
 SUBDIRS:=pkg test
 $(OUTPUT_DIR)/.generated-code:
 	mkdir -p ${OUTPUT_DIR}
-	rm -rf vendor_any
```
gunnar-solo, line 105: since this is now moved to `make clean`, should this target be updated to call `clean`?

Makefile
gunnar-solo, line 18: woo! color-coding!



Screen Shot 2023-02-07 at 9 10 27 AM

Makefile
```diff
@@ -48,14 +69,13 @@ mod-download:
...
 update-deps:
 	mkdir -p $(DEPSGOBIN)
-	GOBIN=$(DEPSGOBIN) go install github.com/solo-io/protoc-gen-ext
-	GOBIN=$(DEPSGOBIN) go install github.com/solo-io/protoc-gen-openapi
-	GOBIN=$(DEPSGOBIN) go install golang.org/x/tools/cmd/goimports
-	GOBIN=$(DEPSGOBIN) go install github.com/golang/protobuf/protoc-gen-go
-	GOBIN=$(DEPSGOBIN) go install github.com/envoyproxy/protoc-gen-validate
-	GOBIN=$(DEPSGOBIN) go install github.com/golang/mock/gomock
-	GOBIN=$(DEPSGOBIN) go install github.com/golang/mock/mockgen
-	GOBIN=$(DEPSGOBIN) go install github.com/onsi/ginkgo/ginkgo
```
gunnar-solo, line 58: explicitly calling out that this is no longer present (intentional, I'm sure).  Specifically, the +/- delta made it a *bit* hard to immediately see

changelog/v0.31.0/ginkgo-v2.yaml
```diff
@@ -0,0 +1,55 @@
...
+    issueLink: https://github.com/solo-io/gloo/issues/7222
+    resolvesIssue: false
+    description: >
+      Upgrade Ginkgo from v1 to v2, using https://onsi.github.io/ginkgo/MIGRATING_TO_V2 as a helpful guide.
+  - type: BREAKING_CHANGE
+    issueLink: https://github.com/solo-io/gloo/issues/7222
+    resolvesIssue: false
+    description: >
+      Remove multicluster package, as it was not used by any projects
+  - type: DEPENDENCY_BUMP
```
gunnar-solo, line 12: how did you choose which dependency bumps to call out, here?

pkg/api/v1/clients/apiclient/resource_client_test.go
```diff
@@ -59,7 +59,11 @@ var _ = Describe("Apiclient", func() {
 
 		port = lis.Addr().(*net.TCPAddr).Port
 		fmt.Fprintf(GinkgoWriter, "grpc listening on %v\n", port)
-		go server.Serve(lis)
+		go func() {
+			defer GinkgoRecover()
```
gunnar-solo, line 63: I haven't seen this pattern, yet.  Can you help me understand what it's doing?

pkg/api/v1/clients/configmap/tests/plain/plain_resource_client_test.go
```diff
@@ -14,20 +15,16 @@ import (
...
 	"k8s.io/client-go/kubernetes"
-
-	// Needed to run tests in GKE
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
 
 var _ = Describe("PlainConfigmap", func() {
 	if os.Getenv("RUN_KUBE_TESTS") != "1" {
-		log.Printf("This test creates kubernetes resources and is disabled by default. To enable, set RUN_KUBE_TESTS=1 in your env.")
+		fmt.Print("This test creates kubernetes resources and is disabled by default. To enable, set RUN_KUBE_TESTS=1 in your env.")
```
gunnar-solo, line 27: per discussion above, I wasn't sure if the goal was to revert this back to `log.`, or leave as-is

pkg/code-generator/codegen/templates/resource_client_test_template.go
```diff
@@ -4,15 +4,16 @@ import (
 	"text/template"
 )
 
-var ResourceClientTestTemplate = template.Must(template.New("resource_client_test").Funcs(Funcs).Parse(`// +build solokit
+var ResourceClientTestTemplate = template.Must(template.New("resource_client_test").Funcs(Funcs).Parse(`
+//go:build solokit
```
gunnar-solo, line 8: only commenting on one of these:  is the new line entirely stylistic?

gunnar-solo (commented):
a few questions, none of which are blocking merge

pkg/utils/kubeutils/metadata.go
jackstine: I believe this has to deal with upgrading `k8s.io/apimachinery` as it is no longer an available property. [This might be of help](https://github.com/kubernetes/apimachinery/blob/fdfff894ab1ea5568846d99cddbca53d876aef77/pkg/apis/meta/v1/types.go#L258).

Makefile
```diff
@@ -147,7 +146,7 @@ GINKGO_VERSION ?= 2.5.0 # match our go.mod
 GINKGO_ENV ?= GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore ACK_GINKGO_DEPRECATIONS=$(GINKGO_VERSION)
 GINKGO_FLAGS ?= -v -tags=purego -compilers=4 --randomize-all --trace -progress -race
 GINKGO_REPORT_FLAGS ?= --json-report=test-report.json --junit-report=junit.xml -output-dir=$(OUTPUT_DIR)
-GINKGO_COVERAGE_FLAGS ?= --cover --covermode=count --coverprofile=coverage.cov
+GINKGO_COVERAGE_FLAGS ?= --cover --covermode=atomic --coverprofile=coverage.cov
```
bewebi, line 149: Why `atomic` rather than `count`?

From [ginkgo docs](https://onsi.github.io/ginkgo/#computing-coverage):
>You can also specify the --covermode to be one of set ("was this code called at all?"), count (how many times was it called?) and atomic (same as count, but threadsafe and expensive). If you run ginkgo --race --cover the --covermode is automatically set to atomic.

Seems like `atomic` should only be needed with `--race`

bewebi: I see now that we do have `-race` specified 👍 🤦

Makefile
```diff
@@ -1,14 +1,36 @@
...
 ROOTDIR := $(shell pwd)
 PACKAGE_PATH:=github.com/solo-io/solo-kit
 OUTPUT_DIR ?= $(ROOTDIR)/_output
 SOURCES := $(shell find . -name "*.go" | grep -v test.go)
 
 GO_BUILD_FLAGS := GO111MODULE=on CGO_ENABLED=0
 
+DEPSGOBIN:=$(OUTPUT_DIR)/.bin
+export PATH:=$(DEPSGOBIN):$(PATH)
+export GOBIN:=$(DEPSGOBIN)
```
jackstine, line 32: maybe some context that these are used in `update-deps` and `generated-code`

Makefile
sam-heilbron: Good call! Will update

Makefile
sam-heilbron: Yeah I always wondered if all of those installs are necessary. I didn't explore the others, but I moved just the installs required for tests (ie installing a particular version of Ginkgo) into its own target.

changelog/v0.31.0/ginkgo-v2.yaml
sam-heilbron: I probably could have added more. I tried to callout internal libraries, or important libraries that affected many files. Are there any deps that you think I should call out?

pkg/api/v1/clients/apiclient/resource_client_test.go
sam-heilbron: Sure! https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure has more details.

```
However, if your spec launches a goroutine that calls Fail (or, equivalently, invokes a failing Gomega assertion), there's no way for Ginkgo to rescue the panic that Fail throws. This will cause the suite to panic and no subsequent specs will run. To get around this you must rescue the panic using defer GinkgoRecover()
```

This change likely wasn't required (as it has lived in our codebase for a while). But I think it's good practice for any goroutines spawned within a test to include the GinkgoRecover.

pkg/api/v1/clients/configmap/tests/plain/plain_resource_client_test.go
sam-heilbron: Ah good call I had missed these, wil update

pkg/api/v1/clients/kube/resource_client.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/solo-io/solo-kit/pkg/utils/specutils"
 
 	"github.com/solo-io/solo-kit/pkg/utils/kubeutils"
-	types "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/types"
```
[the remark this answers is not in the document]
sam-heilbron: I forget. I think I may have done it manually as I was walking through the files, and the IDE nicely showed it grayed out, indicating it wasn't doing anything

pkg/code-generator/codegen/templates/funcs.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/iancoleman/strcase"
-	gendoc "github.com/ilackarms/protoc-gen-doc"
+	gendoc "github.com/pseudomuto/protoc-gen-doc"
```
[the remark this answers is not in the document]
sam-heilbron: Yeah this fork was actually created by Scott W (https://github.com/ilackarms/protoc-gen-doc). As you can see, it forks the repo that I moved us back towards. I tried to parse the differences between the two here, and they seem only tailored to generating code. But from what I could tell we didn't need the fork (created a few years ago) anymore.

It's not ideal (thanks for calling it out). My plan was to largely see if codegen differs (it didn't in solo-kit, but perhaps gloo edge would have more changes) and if there is no diff then move forward with this

pkg/code-generator/codegen/templates/resource_client_test_template.go
sam-heilbron: We're actually converting `// +build solokit` into `//go:build solokit` which is a change in how go1.18 build constraints (https://pkg.go.dev/go/build) are evolved from earlier versions of go

cloudbuild.yaml
```diff
@@ -49,13 +49,9 @@ steps:
 - name: 'gcr.io/$PROJECT_ID/e2e-go-mod-ginkgo:0.4.28'
   entrypoint: 'bash'
   env:
-  - 'ACK_GINKGO_RC=true'
```
sam-heilbron, line 52: These are defined in the Makefile. By specifying necessary variables there, when devs run the make targets locally, they will benefit.

cloudbuild.yaml
```diff
@@ -49,13 +49,9 @@ steps:
...
-  - 'ACK_GINKGO_DEPRECATIONS=1.16.5'
-  - 'GOLANG_PROTOBUF_REGISTRATION_CONFLICT=warn'
   - 'RUN_KUBE_TESTS=1'
   - 'RUN_CONSUL_TESTS=1'
   - 'RUN_VAULT_TESTS=1'
-  - 'TAGGED_VERSION=$TAG_NAME'
```
sam-heilbron, line 58: I just moved the TAGGED_VERSION env var, from being defined on many steps, to being defined at the job level

Makefile
sam-heilbron: I'm happy to. This is a pattern used in all of our repos so I figured a comment might not be necessary since it should be (or become) familiar. What do you think would be a helpful comment?

pkg/code-generator/docs_gen_test.go
```diff
@@ -47,7 +47,11 @@ var _ = Describe("DocsGen", func() {
 		// Create temp directory and path variables
 		workingDir, err := os.Getwd()
 		Expect(err).NotTo(HaveOccurred())
-		tempDir, err = ioutil.TempDir(workingDir, "doc-gen-test-")
+		outputDir := filepath.Join(workingDir, "_output")
```
sam-heilbron, line 50: We previously wrote output files from this test to a `docs-gen-test` directory. I moved this into the `_output` directory so that we can use `make clean` to easily clean these up

Makefile
```diff
@@ -1,14 +1,36 @@
+# https://www.gnu.org/software/make/manual/html_node/Special-Variables.html#Special-Variables
+.DEFAULT_GOAL := help
+
 #----------------------------------------------------------------------------------
-# Base
+# Help
 #----------------------------------------------------------------------------------
+# Our Makefile is quite large, and hard to reason through
```
nfuden, line 7: nit: as we go through adding already large prs lets try to break out smaller concerns like this separately. Given review fatigue we might not check these portions as they should and after the squash there wont be the same history as we might want for things like this.

Makefile
```diff
@@ -1,14 +1,36 @@
...
 ROOTDIR := $(shell pwd)
 PACKAGE_PATH:=github.com/solo-io/solo-kit
 OUTPUT_DIR ?= $(ROOTDIR)/_output
+DEPSGOBIN:=$(OUTPUT_DIR)/.bin
 SOURCES := $(shell find . -name "*.go" | grep -v test.go)
 
 GO_BUILD_FLAGS := GO111MODULE=on CGO_ENABLED=0
 
+export PATH:=$(DEPSGOBIN):$(PATH)
+export GOBIN:=$(DEPSGOBIN)
```
nfuden, line 32: so any and all make commands now go to the output bin? Does this mean that there are no make commands that should ever affect our main go mod?

Makefile
```diff
@@ -101,32 +126,50 @@ generated-code: $(OUTPUT_DIR)/.generated-code update-licenses
...
+	$(GO_BUILD_FLAGS) CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build pkg/api/external/verify.go
 
 #----------------------------------------------------------------------------------
-# Unit Tests
+# Tests
 #----------------------------------------------------------------------------------
 
-# '-skip=multicluster', '-regexScansFilePath' skips any filepath which includes multicluster, which is useful
-# as this code is no longer used
+GINKGO_VERSION ?= 2.5.0 # match our go.mod
```
nfuden, line 145: if it matches our go mod can we pull it from the go mod?

Makefile
jackstine: Important to use binaries built from module.

Makefile
jackstine, line 32:
```suggestion
// Important to use binaries built from module.
export PATH:=$(DEPSGOBIN):$(PATH)
export GOBIN:=$(DEPSGOBIN)
```

test/testutils/locking_suite.go
```diff
@@ -1,54 +1,10 @@
...
 import (
-	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/solo-io/k8s-utils/testutils/clusterlock"
-	"github.com/solo-io/k8s-utils/testutils/kube"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-// call this function with a var _ = in the test suite file
-func LockingSuiteEach() bool {
```
nfuden, line 12: I assume this is safe but can we note why this is a valid removal? I dont see a reference in the description of the pr

Makefile
sam-heilbron: Yeah I couldn't find any make targets that would affect our go mod, also I'm not sure we would want them to any, would we?

pkg/api/v2/reporter/reporter_test.go
```diff
@@ -36,6 +37,7 @@ var _ = Describe("Reporter", func() {
 		mockResourceClient = memory.NewResourceClient(memory.NewInMemoryResourceCache(), &v1.MockResource{})
 		fakeResourceClient = memory.NewResourceClient(memory.NewInMemoryResourceCache(), &v1.FakeResource{})
 		reporter = rep.NewReporter("test", statusClient, mockResourceClient, fakeResourceClient)
+		rep.DisableTruncateStatus = false
```
jackstine, line 40: might be helpful to place a comment as to why `TruncateStatus = false`

Makefile
sam-heilbron: Yeah that's fair, I figured since I was adding targets here it made sense to change them all at once, but I'll try to do that moving forward

test/testutils/locking_suite.go
sam-heilbron: Good call! updated the description

pkg/api/v2/reporter/reporter_test.go
sam-heilbron: I like it, updated

Makefile
jackstine: `go mod graph | grep "github.com/solo-io/solo-projects github.com/onsi/ginkgo" | cut -d '@' -f2 | cut -d 'v' -f2`

sam-heilbron: Happy to. I was lazy and didn't investigate how to do this. Do you have any tips?

sam-heilbron: Yeah so what would be the make syntax? I've been using `go list -m all | grep github.com/onsi/ginkgo/v2 | awk '{print $2}` but I can't set a make variable with that

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=af308736-f078-43bc-a3d7-33be2c32e2d7

successful run, want to see a couple consecutive

gunnar-solo (approved):
Make it so.

Makefile
nfuden: So I thought about this a bit more I have a proposal that is out of scope for this pr as its already too large so ill spin it out into its own issue. Namely we shouldnt be doing dark magic in the make file. We should be checking this with our generate commands that way we can assert all the of the things that change when the go mod changes like we do already with our generate commands in general.

bewebi (approved):

PR #550, opened 2024-02-05
Support validation again snapshot - consistent order of reporter output

sheidkamp:
# Description
* Updated all Validation functions to return errors and warnings in a consistent order
* Added tests

# Context:
`ResourceReports` is defined as `type ResourceReports map[resources.InputResource]Report`

We want to present the results of iterating over these reports in a consistent manner, and the difficulty is that while the `InputResource`s may have unique instantiations, their underlying data might be identical.

The existing implementation does not guarantee any order, so updating to use a specific order will be backwards compatible.

# Details:
The approach taken to ensure output consistency is:
* initialize 2 data structures:
  * `refKeys []string` - a slice of the unique keys
  * `refMap = map[string][]resources.InputResource` - a map of the keys to the InputResources that are identified by that key.
* loop over the ResourceReports and for each key `res`, which is an InputResource:
  * `resKey := res.GetMetadata().Ref().String()` as the key for the resource 
  * add the `resKey` to `refKeys`
  * add the resource `res` to the slice of resources at `refMap[resKey]`
* Sort the `refKeys` array

We can now loop over `refKeys` and access the resources in a consistent order. We still need to handle the ordering of resources that share a key, and we will do that in the individual validation functions.

In the validation functions:
* loop over the sorted keys in `refKeys`
* for each key, loop over the associated resources
  * for each resource, look up the report and append the errors/warnings to a slice
  * after processing the reports for a given key, there will be a slice of errors
    * this is a 1d slice, as the errors from each report will be contained in a single multierror
  * sort this array
  * append the elements of the sorted array of errors to the running error.

By executing this process, we can say for a given set of reports:
* The resources will be iterated over in groups ordered by `res.GetMetadata().Ref().String()`
* Within the groups of the resources the sets of errors will sorted
* Because of this the sorting will be consistent each time Validation is run

## Interesting decisions
There is an existing idiosyncrasy of the error reporting where an "invalid resource" error is added only the first resource encountered with an error: https://github.com/solo-io/solo-kit/blob/2f0d239abcbc7ef5a8fdc756a3a68e27eb0f9ab9/pkg/api/v2/reporter/reporter.go#L201 . This means if there are multiple resources with errors, only one will generate an  "invalid resource" error.

This behavior has been maintained in the interest of rapid development

### DIFF
--- a/pkg/api/v2/reporter/reporter.go
+++ b/pkg/api/v2/reporter/reporter.go
@@ -246,7 +246,7 @@ func (e ResourceReports) Validate() error {
 		// name/namespace is not unique, so we collect those references together
 		reses := refMap[refKey]
 
-		errsForKey := []error{}
+		var errsForKey []error
 		for _, res := range reses {
 			rpt := e[res]
 
@@ -277,7 +277,7 @@ func (e ResourceReports) ValidateStrict() error {
 	refMap, refKeys := e.refMapAndSortedKeys()
 
 	for _, refKey := range refKeys {
-		errsForKey := []error{}
+		var errsForKey []error
 		reses := refMap[refKey]
 
 		// name/namespace is not unique, so we collect those references together
@@ -317,7 +317,7 @@ func (e ResourceReports) ValidateSeparateWarnings() (error, error) {
 
 	for _, refKey := range refKeys {
 		// name/namespace is not unique, so we collect those references together
-		warnForKey := []error{}
+		var warnForKey []error
 		reses := refMap[refKey]
 
 		for _, res := range reses {
--- a/pkg/api/v2/reporter/reporter.go
+++ b/pkg/api/v2/reporter/reporter.go
@@ -3,6 +3,7 @@ package reporter
 import (
 	"context"
 	"os"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/api/v2/reporter/reporter.go
```diff
@@ -192,38 +193,180 @@ func (e ResourceReports) FilterByKind(kind string) ResourceReports {
...
+			if rpt.Errors != nil {
+				errsForKey = append(errsForKey, rpt.Errors)
+			}
+		}
+
+		if len(errsForKey) > 0 {
 			if errs == nil {
-				errs = errors.Errorf("invalid resource %v.%v", res.GetMetadata().Namespace, res.GetMetadata().Name)
+				// All of the resources in the group have the same metadata, so use the first one
+				errs = errors.Errorf("invalid resource %v.%v", reses[0].GetMetadata().Namespace, reses[0].GetMetadata().Name)
```
davidjumani, line 261: nit: I understand that this is the way things have been, but this can be confusing as if there are multiple resources and errors in some of them, it will mention only the first resource and append errors belonging to other resources

pkg/api/v2/reporter/reporter.go
```diff
@@ -192,38 +193,180 @@ func (e ResourceReports) FilterByKind(kind string) ResourceReports {
...
+		for _, res := range reses {
+			rpt := e[res]
+			if len(rpt.Warnings) > 0 {
+				errsForKey = append(errsForKey, errors.Errorf("WARN: \n  %v", rpt.Warnings))
+
+			}
+		}
+
+		if len(errsForKey) > 0 {
 			if errs == nil {
```
davidjumani, line 293: Same here

davidjumani (approved):
Couple of nits. Apart from that LGTM

pkg/api/v2/reporter/reporter.go
```diff
@@ -192,38 +193,180 @@ func (e ResourceReports) FilterByKind(kind string) ResourceReports {
...
+	// With Strict WarningHandling, warnings are treated as errors
+	Strict WarningHandling = iota
+	// With IgnoreWarnings WarningHandling, warnings are ignored
+	IgnoreWarnings
+	// With SeparateWarnings WarningHandling, warnings are returned separately from errors
+	SeparateWarnings
+)
+
+// ValidateReport validates the reports according to the given validation type.
+func (e ResourceReports) ValidateWithWarnings(t WarningHandling) (error, error) {
```
davidjumani, line 357: nit: Good place for named returns

pkg/api/v2/reporter/reporter.go
```diff
@@ -192,34 +193,118 @@ func (e ResourceReports) FilterByKind(kind string) ResourceReports {
...
+	var refMap = make(map[string][]resources.InputResource)
+
+	// Loop over the resources
+	for res := range e {
+
+		// Get a string representation of the resource ref. This is not guaranteed to be unique.
+		resKey := res.GetMetadata().Ref().String()
+
+		// Add the key to the list of keys if it is not already there
+		if !slices.Contains(refKeys, resKey) {
```
davidjumani, line 216: nit: Instead of keeping the refKeys and the map, and checking if the key exists in the refKeys before adding it, would it just be simpler to get the refKeys after the map has been populated ?

pkg/api/v2/reporter/reporter.go
sheidkamp: Removed, so no longer relevant

pkg/api/v2/reporter/reporter.go
```diff
@@ -192,34 +193,122 @@ func (e ResourceReports) FilterByKind(kind string) ResourceReports {
...
+// The previous implementation did not guarantee a consistent order when iterating over the resources,
+// so any order used here will be acceptable for backwards compatibility.
+func (e ResourceReports) refMapAndSortedKeys() (map[string][]resources.InputResource, []string) {
+	// refKeys contains all the unique keys for the resources
+	var refKeys []string
+	// refMaps is a map of resource keys to a slice of resources with that key
+	var refMap = make(map[string][]resources.InputResource)
+
+	// Loop over the resources
+	for res := range e {
```
ben-taussig-solo, line 210: nit: I suspect that you considered this approach, but we can remove the second loop if we do something like:

```go
		// Get a string representation of the resource ref. This is not guaranteed to be unique.
		resKey := res.GetMetadata().Ref().String()

		if _, exists := refMap[resKey]; !exists {
		    refKeys = append(refKeys, resKey)
		}

		// Add the resource to the map of resources with the same key
		refMap[resKey] = append(refMap[resKey], res)
```
That being said, I don't know a lot about the performance of slices in golang so if the current approach (i.e. creating refMap after we know the number of refKeys) is meaningfully faster then we should leave it

ben-taussig-solo: After reviewing the rest of the PR, this is my only comment. Given that this code is used in the control-plane (i.e. is not performance critical) and I expect the length of refKeys to essentially never be larger than ~1000 (in the case of the largest customer environments), I am approving the PR.

ben-taussig-solo (approved):

pkg/api/v2/reporter/reporter.go
sheidkamp: Lol, had this approach, then backed it out this morning based on David's comment: https://github.com/solo-io/solo-kit/pull/550/files/412405aac0b1d3498c4ee350e8992360f3045dba#r1496090964

PR #551, opened 2024-05-21
in-memory resource client: support expression selector

jenshu:


pkg/api/v1/clients/memory/resource_client.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
 	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
 	"github.com/solo-io/solo-kit/pkg/errors"
+	"github.com/solo-io/solo-kit/pkg/utils/kubeutils"
```
sam-heilbron, line 15: This feels a bit awkward. Our solo-kit library attempts to separate out the persistence layers intentionally, and this seems to blend the kubernetes and in-memory implementations. Is there a way around this, or a side-effect of using types in our in-memory client that are kubernetes specific?

pkg/api/v1/clients/memory/resource_client_test.go
```diff
@@ -62,6 +62,178 @@ var _ = Describe("Base", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(resources).To(HaveLen(1))
+			Expect(resources[0].GetMetadata().GetName()).To(Equal("name4"))
+		})
+
+		It("when both selector and expression selector are provided, uses expression selector", func() {
+			resources, err := client.List("ns2", clients.ListOpts{
+				Selector: map[string]string{
+					"ignored": "selector",
```
davidjumani, line 226:
```suggestion
					"key": "val2",
```
To ensure that the selector is ignored

pkg/api/v1/clients/memory/resource_client.go
sam-heilbron: Though I guess the below `labels` import already broke this paradigm, so we can ignore my earlier comment

davidjumani (dismissed):

pkg/api/v1/clients/memory/resource_client.go
```diff
@@ -197,9 +198,15 @@ func (rc *ResourceClient) Delete(namespace, name string, opts clients.DeleteOpts
 func (rc *ResourceClient) List(namespace string, opts clients.ListOpts) (resources.ResourceList, error) {
 	opts = opts.WithDefaults()
 	cachedResources := rc.cache.List(rc.Prefix(namespace))
+
+	labelSelector, err := kubeutils.ToLabelSelector(opts)
```
sam-heilbron, line 202: Should we gate this functionality to only apply it _if_ a labelSelector is non-nil?

changelog/v0.34.3/mem-client-expression-selector.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NON_USER_FACING
```
sam-heilbron, line 2: It feels like since this is a library, this is user facing, in that we are now respecting ListOpts and WatchOpts on the in memory resource client?

pkg/api/v1/clients/memory/resource_client_test.go
```diff
@@ -62,6 +62,178 @@ var _ = Describe("Base", func() {
...
+
+			// resources are sorted by namespace, then name
+			expectedResourceNames := []string{
+				"name1", "name3", "name5", // ns1
+				"name2", "name4", // ns2
+			}
+			Expect(resources).To(HaveLen(len(expectedResourceNames)))
+			for i, r := range resources {
+				Expect(r.GetMetadata().GetName()).To(Equal(expectedResourceNames[i]))
+			}
```
sam-heilbron, line 145: nit: Any reason we don't do something like:
```suggestion
			Expect(resources).To(HaveLen(len(expectedResourceNames)))
			Expect(resources).To(ContainElements(obj1, obj3, obj5, obj2, obj4))
```

We could have a standalone test for proper ordering, but it feels like this way we could remove the need to refine all the names and just reference the objects directly

pkg/api/v1/clients/memory/resource_client_test.go
```diff
@@ -62,6 +62,178 @@ var _ = Describe("Base", func() {
...
+					"ignored": "selector",
+				},
+				ExpressionSelector: "key in (val1,val3)",
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(resources).To(HaveLen(1))
+			Expect(resources[0].GetMetadata().GetName()).To(Equal("name4"))
+		})
+	})
```
sam-heilbron, line 235: Do you think it makes sense to extend the benchmark tests below to include the selector and expression selector? As this memroy resource client becomes used more widely in our product, it feels like it could be valuable

changelog/v0.34.3/mem-client-expression-selector.yaml
jenshu: heh yeah, i saw that a lot of other 'user-facing' changes in solo-kit had NON_USER_FACING changelogs, so i went with that. but i can file an issue and link it here

pkg/api/v1/clients/memory/resource_client_test.go
jenshu: do you mind if we keep as-is for now? currently, the resources returned from the List call contain extra metadata and so i can't directly use ContainElements, need to define our own matcher. it's taking me a bit to get it working correctly so maybe we can save that for a follow-up?

pkg/api/v1/clients/memory/resource_client.go
jenshu: a04fd4046965b63a9381227c3dece888d20ed542

changelog/v0.34.3/mem-client-expression-selector.yaml
jenshu: 4129802

pkg/api/v1/clients/memory/resource_client_test.go
jenshu: added benchmark tests 3fad32cc61a62e2e02d441c92f16d2e076e967d8
i realized that the existing test wasn't being run (at least not when i ran it locally), probably because `Measure` is deprecated, so i updated to use gmeasure

sam-heilbron (approved):

davidjumani (approved):

PR #552, opened 2024-05-21
chore: bump k8s to 1.29

davidjumani:
Bump the following to support k8s v1.29 :
- k8s
- k8s-utils 
- cloudbuilder

Ref: https://github.com/solo-io/gloo/issues/8978


changelog/v0.35.0/bump-k8s-129.yaml
```diff
@@ -0,0 +1,54 @@
...
+  dependencyRepo: zap
+  dependencyTag: v1.26.0
+- type: DEPENDENCY_BUMP
+  dependencyOwner: solo-io
+  dependencyRepo: k8s-utils
+  dependencyTag: v0.7.2
+- type: DEPENDENCY_BUMP
+  dependencyOwner: solo-io
+  dependencyRepo: cloud-builders
+  dependencyTag: v0.9.1
```
sheidkamp, line 54: Last time we updated we also updated `go-utils` (https://github.com/solo-io/solo-kit/blob/e7296cef16898146ed5eedba199c54df2f47ede0/changelog/v0.34.0/bump-k8s-128.yaml#L39) which looks like its still on 1.28. Do we need to bump that as well?

sheidkamp (commented):
One question

changelog/v0.35.0/bump-k8s-129.yaml
davidjumani: go-utils does not import any k8s libs so there's no need to update it. The last bump only updated some other libraries used by k8s but it isn't necessary to update them

sheidkamp (dismissed):

nfuden (dismissed):

jenshu (approved):

PR #554, opened 2024-05-28
implement json.Marshaler in snapshot templates

jbohanon:
Adds explicit implementation of `json.Marshaler` to the snapshots template.

The impetus for this is to enable returning snapshots from an admin interface in Gloo https://github.com/solo-io/gloo/pull/9517


pkg/code-generator/codegen/templates/snapshot_template.go
```diff
@@ -83,6 +88,10 @@ func (s {{ .GoName }}Snapshot) HashFields() []zap.Field {
 	return append(fields, zap.Uint64("snapshotHash",  snapshotHash))
 }
 
+func (s {{ .GoName }}Snapshot) MarshalJSON() ([]byte, error) {
+    return json.Marshal(&s)
```
nfuden, line 92: Do we need to do something like https://github.com/solo-io/skv2/blob/2de765daa5b28dd72309cdcac2477ebd4b2ea84a/contrib/codegen/templates/input/input_snapshot.gotmpl#L302

nfuden (approved):

pkg/code-generator/codegen/templates/snapshot_template.go
jbohanon: Slack conversation:
```log
Nathan Fudenberg
Should we mimick skv2 redact secrets? https://github.com/solo-io/solo-kit/pull/554/files#r1617533556


Jacob Bohanon
It's a bit impractical as of now since our secrets are defined in gloo, so we would have to put in a solo-apis or gloo dep in to get the structs or recurse with string object names which is even more fragile


Nathan Fudenberg
makes sense
```

davidjumani, line 92: Would the `&` cause any nil issues ?

davidjumani (approved):

PR #555, opened 2024-06-14
Add option to disable kubebuilder markers

arianaw66:
# Description

Adds option to disable kubebuilder markers and validations to omit them from the generated OpenAPI schema.

Changes can be seen in-action [here](https://github.com/solo-io/gloo/pull/9481/commits/1ccc94ea1d7055a097fd9bb8a6c9f50003fef3a5)

# Context

This is needed for https://github.com/solo-io/gloo-mesh-enterprise/issues/17005

When disableKubeMarkers is set to true, `Type=object` annotations will be respected such that recursive fields will still work. `Required` annotations will also be respected.

This is essentially a cherry-pick of https://github.com/solo-io/skv2/pull/567.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	github.com/solo-io/go-utils v0.24.8
 	github.com/solo-io/k8s-utils v0.7.2
 	github.com/solo-io/protoc-gen-ext v0.0.18
-	github.com/solo-io/protoc-gen-openapi v0.1.1
+	github.com/solo-io/protoc-gen-openapi v0.2.4
 	go.opencensus.io v0.24.0
 	go.uber.org/zap v1.26.0
 	golang.org/x/sync v0.5.0
@@ -97,7 +97,7 @@ require (
 	github.com/lyft/protoc-gen-star/v2 v2.0.3 // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
-	github.com/mattn/go-isatty v0.0.17 // indirect
+	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-zglob v0.0.1 // indirect
 	github.com/mitchellh/copystructure v1.2.0 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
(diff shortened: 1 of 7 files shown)


shashankram (approved):

marcogschmidt (approved):

go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	github.com/solo-io/go-utils v0.24.8
 	github.com/solo-io/k8s-utils v0.7.2
 	github.com/solo-io/protoc-gen-ext v0.0.18
-	github.com/solo-io/protoc-gen-openapi v0.1.1
+	github.com/solo-io/protoc-gen-openapi v0.2.4
```
sam-heilbron, line 35: nit: When bumping libraries across minor versions, I tend to callout in the PR body what those changes were. Mostly just to get in the habit of validating changes, because we have noticed that sometimes there are behavioral change that impact us. Since this is an internal library that we manage, and is low risk, it's certainly not a requirement (more sharing as it's a pattern I follow)

pkg/code-generator/collector/executor.go
```diff
@@ -105,10 +109,11 @@ func (o *OpenApiProtocExecutor) Execute(protoFile string, toFile string, imports
 	_ = os.Mkdir(directoryPath, os.ModePerm)
 
 	cmd.Args = append(cmd.Args,
-		fmt.Sprintf("--openapi_out=yaml=true,single_file=false,include_description=%v,enum_as_int_or_string=%v,additional_empty_schema=%v:%s",
+		fmt.Sprintf("--openapi_out=yaml=true,single_file=false,include_description=%v,enum_as_int_or_string=%v,additional_empty_schema=%v,disable_kube_markers=%v:%s",
```
sam-heilbron, line 112: I like this change, because it follows the precedent that existed. An alternative way to perhaps improve this "API" would be to just expose a raw string to users. This string would be passed from callers of the code, down to protoc. This way, the next time we add an argument to our open-api protoc impelementation, we don't need to pipe through values, and can instead just update the string

sam-heilbron (commented):
nice work!

pkg/code-generator/collector/executor.go
arianaw66: I'm not sure I prefer that to exposing a clear option in solo-kit that IDEs won't let you typo and such, but certainly a conversation we can have.

PR #558, opened 2024-09-03
Add RemoveMatches in snapshot template

davidjumani:
- Adds the `RemoveMatches` method in `snapshot_template.go` that removes all resources that match the given crietria.
- Adds the ability for event loops to determine if the Syncer passed is a `SyncDecider` to determine if a sync is necessary.


pkg/code-generator/codegen/templates/event_loop_template.go
```diff
@@ -92,16 +92,33 @@ func (el *{{ lower_camel .GoName }}EventLoop) Run(namespaces []string, opts clie
...
+				if syncDecider, isDecider := el.syncer.({{ .GoName }}SyncDecider); isDecider {
+					if shouldSync := syncDecider.ShouldSync(previousSnapshot, snapshot); !shouldSync {
+						continue // skip syncing this syncer
+					}
+				} else if syncDeciderWithContext, isDecider := el.syncer.({{ .GoName }}SyncDeciderWithContext); isDecider {
+					if shouldSync := syncDeciderWithContext.ShouldSync(opts.Ctx, previousSnapshot, snapshot); !shouldSync {
+						continue // skip syncing this syncer
+					}
+				}
+
```
davidjumani, line 121: Copied over from https://github.com/solo-io/solo-kit/blob/24798b5ad0c13f00d3d91ffaf3ff892d84c956e6/pkg/code-generator/codegen/templates/simple_event_loop_template.go#L90

test/mocks/v1alpha1/testing_event_loop.sk.go
```diff
@@ -87,16 +87,33 @@ func (el *testingEventLoop) Run(namespaces []string, opts clients.WatchOpts) (<-
...
+		var previousSnapshot *TestingSnapshot
+
 		for {
 			select {
 			case snapshot, ok := <-watch:
 				if !ok {
 					return
 				}
+
+				if syncDecider, isDecider := el.syncer.(TestingSyncDecider); isDecider {
```
sam-heilbron, line 107: When this is running, how should I expect to debug this behavior? Should we have metrics or logs emitted when a sync is not necessary?

test/mocks/v1/testing_snapshot.sk.go
```diff
@@ -257,6 +257,65 @@ func (s *TestingSnapshot) RemoveFromResourceList(resource resources.Resource) er
 	}
 }
 
+func (s *TestingSnapshot) RemoveAllResourcesInNamespace(namespace string) {
```
sam-heilbron, line 260: Should we store resources indexed by namespace so that we can more easily find/delete them?

test/mocks/v1/testing_snapshot.sk.go
```diff
@@ -257,6 +257,65 @@ func (s *TestingSnapshot) RemoveFromResourceList(resource resources.Resource) er
...
+	var Simplemocks SimpleMockResourceList
```
sam-heilbron, line 261: It feels like this is a specific implementation of a Map/Reduce type function. What if instead of just accepting a namespace, we have a function
```suggestion
	func (s *TestingSnapshot) RemoveMatches(predicate func(core.Metadata) {
```

And then the predicate we pass in is a function that inspects the metadata.namespace?

test/mocks/v1alpha1/testing_event_loop.sk.go
davidjumani: Yes, the implementation can have metrics to track this

test/mocks/v1/testing_snapshot.sk.go
davidjumani: Yeah this is a better idea. Let me update this

pkg/code-generator/codegen/templates/event_loop_template.go
```diff
@@ -92,16 +92,33 @@ func (el *{{ lower_camel .GoName }}EventLoop) Run(namespaces []string, opts clie
...
 		for {
 			select {
 			case snapshot, ok := <-watch:
 				if !ok {
 					return
 				}
+
+				if syncDecider, isDecider := el.syncer.({{ .GoName }}SyncDecider); isDecider {
+					if shouldSync := syncDecider.ShouldSync(previousSnapshot, snapshot); !shouldSync {
```
jbohanon, line 113: nit:
```suggestion
					if !syncDecider.ShouldSync(previousSnapshot, snapshot) {
```

pkg/code-generator/codegen/templates/event_loop_template.go
```diff
@@ -92,16 +92,33 @@ func (el *{{ lower_camel .GoName }}EventLoop) Run(namespaces []string, opts clie
...
+						continue // skip syncing this syncer
+					}
+				} else if syncDeciderWithContext, isDecider := el.syncer.({{ .GoName }}SyncDeciderWithContext); isDecider {
+					if shouldSync := syncDeciderWithContext.ShouldSync(opts.Ctx, previousSnapshot, snapshot); !shouldSync {
```
jbohanon, line 117: nit:
```suggestion
					if !syncDeciderWithContext.ShouldSync(opts.Ctx, previousSnapshot, snapshot) {
```

jbohanon (approved):

test/mocks/v1alpha1/testing_event_loop.sk.go
sam-heilbron: Is that something that we need to implement here, or in Gloo we can define that SyncDecider?

sam-heilbron (approved):

PR #562, opened 2024-09-19
Fix duplicate definition of predicate

davidjumani:
Fixes the issue of multiple declarations of Predicate when multiple snapshots are generated

This causes compilation issues after running codegen on the latest v0.35.x

### DIFF
--- a/changelog/v0.35.4/fix-duplicate-predicate.yaml
+++ b/changelog/v0.35.4/fix-duplicate-predicate.yaml
@@ -1,0 +1,6 @@
+changelog:
+- type: FIX
+  issueLink: https://github.com/solo-io/gloo/issues/9274
+  resolvesIssue: false
+  description: Fixes the issue of multiple declarations of Predicate when multiple snapshots are generated
+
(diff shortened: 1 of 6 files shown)


pkg/api/v1/resources/core/metadata_extensions.go
```diff
@@ -17,3 +17,5 @@ func (m *Metadata) Ref() *ResourceRef {
 func (m *Metadata) Match(ref *ResourceRef) bool {
 	return m.GetNamespace() == ref.GetNamespace() && m.GetName() == ref.GetName()
 }
+
+type Predicate func(metadata *Metadata) bool
```
sam-heilbron, line 21: nit: Since this is in the `core` package, `core.Predicate` is kind of generic thing. Perhaps we clarify this is a `MetadataPredicate`?

changelog/v0.35.4/fix-duplicate-predicate.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+- type: FIX
+  issueLink: https://github.com/solo-io/gloo/issues/9274
+  resolvesIssue: false
+  description: Fixes the issue of multiple declarations of Predicate when multiple snapshots are generated
```
sam-heilbron, line 5: I agree with @nfuden 's point in Slack, could we either inclue the error we see in the PR body, or add a test to show that we can use this Predicate (which would prove that it would work in gloo as well)?

sam-heilbron (approved):
LGTM!

PR #563, opened 2024-09-19
Fix duplicate definition of predicate

davidjumani:
Fixes the issue of multiple declarations of Predicate when multiple snapshots are generated

This causes compilation issues after running codegen on latest main

Eg: In gloo OSS when multiple snapshots are created within the same package (discovery snapshot, setup snapshot), multiple `Predicate` definitions are generated as this is in the _template file.
This moves it to the core package to resolve this issue


sam-heilbron (approved):

timflannagan:
Just ran into this locally with a solo-kit replace pin pointing to my local copy. LGTM, but I'm going to pull this down locally and verify it fixes everything first.

timflannagan (approved):

PR #567, opened 2024-11-14
Update template_funcs.go

sheidkamp:
We are removing external API documentation from our docs, so we need to stop linking to them when generating docs from protobufs.

The current solution is to no longer render the link, as there are complications in programmatically determining the correct externally hosted documentation location and version.

Discussion with docs: https://solo-io-corp.slack.com/archives/C03MFATU265/p1731608037978029?thread_ts=1731598700.372589&cid=C03MFATU265

Example change when pulled into `gloo`:
before:
```
| `responseTransformation` | [.envoy.api.v2.filter.http.TransformationTemplate](../../../../external/envoy/extensions/transformation/transformation.proto.sk/#transformationtemplate) |  |
```

after
```
| `responseTransformation` | .envoy.api.v2.filter.http.TransformationTemplate |  |
```

Example usage: https://github.com/solo-io/gloo/pull/10065

### DIFF
--- a/pkg/code-generator/docgen/funcs/template_funcs.go
+++ b/pkg/code-generator/docgen/funcs/template_funcs.go
@@ -319,9 +319,11 @@ func linkForField(project *model.Project, docsOptions *options.DocsOptions) func
 				//return "", errors.Errorf("failed to get generated file path for proto %v in list %v", file.GetName(), project.Request.FileToGenerate)
 			}
 
-			// Don't link to external apis docs, as they will be deleted
-			if strings.HasPrefix(linkedFile, "github.com/solo-io/gloo/projects/gloo/api/external") {
-				return typeName, nil
+			// Skip links for packages that are configured to be skipped
+			for _, pkg := range docsOptions.RenderOptions.GetSkipLinksForPackages() {
+				if strings.HasPrefix(linkedFile, pkg) {
+					return typeName, nil
+				}
 			}
 
 			linkedFile = relativeFilename(forFile.GetName(), linkedFile)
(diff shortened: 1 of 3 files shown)


davidjumani (dismissed):

pkg/code-generator/docgen/funcs/template_funcs.go
```diff
@@ -318,8 +318,13 @@ func linkForField(project *model.Project, docsOptions *options.DocsOptions) func
 				linkedFile = filepath.Base(file.GetName())
 				//return "", errors.Errorf("failed to get generated file path for proto %v in list %v", file.GetName(), project.Request.FileToGenerate)
 			}
-			linkedFile = relativeFilename(forFile.GetName(), linkedFile)
 
+			// Don't link to external apis docs, as they will be deleted
+			if strings.HasPrefix(linkedFile, "github.com/solo-io/gloo/projects/gloo/api/external") {
```
jenshu, line 323: is there a way to make this non-gloo-specific? e.g. pass in the repo name to the func?

sheidkamp: good idea, implemented, and wrapped it up in a struct with this issue in mind: https://github.com/k8sgateway/k8sgateway/issues/10349

pkg/code-generator/docgen/options/options.go
```diff
@@ -21,8 +21,21 @@ type HugoOptions struct {
...
-	Output      DocsOutput
-	HugoOptions *HugoOptions
+	Output        DocsOutput
+	HugoOptions   *HugoOptions
+	RenderOptions *RenderOptions
+}
+
+// RenderOptions provides options for rendering documentation
+type RenderOptions struct {
+	SkipLinksForPackages []string // when rendering markdown, do not attempt to link to these packages or their subpackages
```
jenshu, line 31: is this a file path prefix? can you put in the comments here an example of what one of these strings would look like?

sheidkamp: renamed and updated

jenshu (approved):

davidjumani (approved):

PR #571, opened 2024-11-16
docs: update markdown template 

sam-heilbron:
# Background
https://github.com/solo-io/solo-projects/issues/6612#issuecomment-2447651284 contains the necessary context behind these changes.

# Testing
Testing this locally was difficult, and it was easier to just run code generation in Gloo. I opened https://github.com/solo-io/solo-kit/issues/572 to track this gap in local functionality.

I opened https://github.com/solo-io/gloo/pull/10362 to demonstrate how this would impact the generated docs in Gloo


Nadine2016 (dismissed):

pkg/code-generator/schemagen/crd.go
```diff
@@ -22,6 +22,8 @@ const (
 	v1 = "apiextensions.k8s.io/v1"
 )
 
+const crdFileHeader = `CRD ValidationSchema generated by solo-kit. DO NOT EDIT.`
```
arianaw66, line 25: Is `ValidationSchema` meaningful? The only instance of that string I see in the `gloo` repo is the `ValidationSchemaOptions`. Can we change it or just make it "validation schema"?

pkg/code-generator/docgen/generator.go
```diff
@@ -147,22 +147,37 @@ func (d *DocsGen) FileHeader(filename string) string {
...
+	name := filepath.Base(filename)
+
+	var title string
+	if strings.HasSuffix(name, d.protoSuffix()) {
+		// Remove the "sk.md" extensions
+		name = name[:len(name)-len(d.protoSuffix())]
+
+		// Remove the ".proto" extension
+		extension := filepath.Ext(name)
+		title = strcase.ToCamel(name[0 : len(name)-len(extension)])
```
arianaw66, line 166: Nit: let's keep it consistent with the other line that removes the extension
```suggestion
		title = strcase.ToCamel(name[:len(name)-len(extension)])
```

arianaw66, line 166: Nit: let's keep it consistent with the other line that removes the extension
```suggestion
		title = strcase.ToCamel(name[:len(name)-len(extension)])
```

pkg/code-generator/docgen/generator.go
```diff
@@ -147,22 +147,37 @@ func (d *DocsGen) FileHeader(filename string) string {
...
+	} else {
+		// Not a file for a proto file, leave the title to match the name of the file
```
arianaw66, line 168: Am I understanding this right?
```suggestion
		// Not a file generated by a proto file, leave the title to match the name of the file
```

pkg/code-generator/docgen/generator.go
arianaw66, line 166: If it's always the same ".proto.sk.md" extension, why not do this in one modification?
```suggestion
        protoExtension := ".proto"+d.protoSuffix()
	if strings.HasSuffix(name, protoExtension) {
		// Remove the "proto.sk.md" extension
		name = name[:len(name)-len(protoExtension)]

		title = strcase.ToCamel(name)
```

pkg/code-generator/docgen/generator.go
```diff
@@ -147,22 +147,37 @@ func (d *DocsGen) FileHeader(filename string) string {
...
+`
+}
+
+func (d *DocsGen) hugoFileHeader(filename string) string {
+	name := filepath.Base(filename)
+
+	var title string
+	if strings.HasSuffix(name, d.protoSuffix()) {
+		// Remove the "sk.md" extensions
+		name = name[:len(name)-len(d.protoSuffix())]
```
arianaw66, line 162: Out of scope, but why not this?
```suggestion
		name = strings.TrimSuffix(name, d.protoSuffix())
```

pkg/code-generator/docgen/generator.go
arianaw66, line 166: I noticed in the `gloo` PR this results in these titles: `GatewaySoloIo`, `EnvoyGlooeSoloIo`
Is that what we want?
Perhaps we only want to camelCase names that have no `.` remaining after stripping the extension?

Nadine2016: so I agree with @arianaw66  that in this case the name does not make sense. Only the ones that end in `.proto` I think should get the name change. the ones that end in `.project` should not. However, I also think we should not have any of these .project resources in our docs, especially since those are just overview pages with all the CRs. So what we could do is one of the following: 

Option 1: 
- Not change the name for .project files. 
- Display them as-is in the docs. 

Option 2: 
- Leave the change as-is and remove them from the docs. We can do this manually on our end if it can't be done in the template or is too hard to implement.

arianaw66 (approved):
LGTM provided we're okay with Option 1 and re-adding `.sk.md` in the docs titles for .project files per https://solo-io-corp.slack.com/archives/C03MFATU265/p1732053402150619?thread_ts=1731948029.520829&cid=C03MFATU265

Nadine2016 (approved):

arianaw66 (approved):
Thanks for the helpful comments!

PR #580, opened 2025-06-12
Bump to k8s 1.33 and other dependencies

ymesika:
Part of https://github.com/solo-io/solo-projects/issues/8102


go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/solo-io/solo-kit
 
-go 1.24
+go 1.24.4
```
davidjumani, line 3: Do we want to pin it to a patch version ?

PR #584, opened 2025-09-22
anchors in headings

artberger:
followup to https://github.com/solo-io/solo-kit/pull/583, for https://github.com/solo-io/doctopus/issues/124, more context in [Slack](https://solo-io-corp.slack.com/archives/CE0R6KPS5/p1758298278068919)

Tested by doing this in the `gloo` repo:
1. Pull in my changes in this PR by running `go get github.com/solo-io/solo-kit@adb-anchors2`
2. `make generate-all -B`
3. example of changes in this commit: https://github.com/solo-io/gloo/commit/1c5105e992a09d6e1703164a8bc0f43d777a604c
BOT NOTES: 
resolves https://github.com/solo-io/doctopus/issues/124

### DIFF
--- a/pkg/code-generator/docgen/funcs/template_funcs.go
+++ b/pkg/code-generator/docgen/funcs/template_funcs.go
@@ -135,7 +135,18 @@ func TemplateFuncs(project *model.Project, docsOptions *options.DocsOptions) tem
 
 func toHeading(docsOptions *options.DocsOptions) func(format string, p *string) string {
 	if docsOptions.Output == options.Hugo {
-		return printPointer
+		// For Hugo, we need to generate the same numbered anchors as toAnchorLink
+		// Track anchor names to handle duplicates
+		anchorCounts := make(map[string]int)
+		return func(format string, p *string) string {
+			val := printPointer(format, p)
+			name := strings.ToLower(val)
+			anchorCounts[name]++
+			if anchorCounts[name] > 1 {
+				return fmt.Sprintf("%s {#%s-%d}", val, name, anchorCounts[name]-1)
+			}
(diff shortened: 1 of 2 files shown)


pkg/code-generator/docgen/funcs/template_funcs.go
```diff
@@ -135,7 +135,18 @@ func TemplateFuncs(project *model.Project, docsOptions *options.DocsOptions) tem
 
 func toHeading(docsOptions *options.DocsOptions) func(format string, p *string) string {
 	if docsOptions.Output == options.Hugo {
-		return printPointer
+		// For Hugo, we need to generate the same numbered anchors as toAnchorLink
+		// Track anchor names to handle duplicates
+		anchorCounts := make(map[string]int)
```
andy-fong, line 140: At first, this looks a bit strange but is the idea that anchorCounts is declare here because toHeading() is called only once and anchorCounts get "auto"captured into the returned function and that's the function that will get called multiple times? That's a clever hack without local static like in C++.

andy-fong (approved):

pkg/code-generator/docgen/funcs/template_funcs.go
artberger: yeah the toHeading is called once during the template setup, but this returned function is called for each heading instance. We needed separate anchorCounts funcs under both toHeading and toAnchorLink because they were counting different things (headings vs anchors), so to keep them in sync so the headings match the anchor counts. Before without this, it was only counting the anchors and replacing it for non-heading links 😅

Nadine2016 (approved):

PR #586, opened 2026-01-30
don't remove xds watch if we didn't respond

andy-fong:
This is a PR to illustrate the problem. Once we agree on the proper fix, I will clean this up with proper changelog.


jenshu (approved):

andy-fong:
Talked to Yuval and since this is not critical for the fix, prefer not to touch solo-kit. So, closing this PR.

PR #588, opened 2026-03-16
chore: Bump k8s to 1.35 and go to 1.26

davidjumani:


puertomontt (dismissed):

nfuden:
Looks like kicks probably wont work as something about type inference is failing after updates.

See the failing line 
Step #7 - "test": {"level":"warn","ts":"2026-03-16T19:56:50.409Z","caller":"configmap/converters.go:105","msg":"invalid resource type (*v1.MockResource) used for plain configmap. unable to convert to kube configmap. only resources with fields of type string are supported for the plain configmap client."}

davidjumani:
Thanks @nfuden I guess the output is cached, so I don't see anything past 
```
Operation completed over 1 objects/514.9 KiB.                                    
tar: Substituting `.' for empty member name
```

nfuden:
Yeah so if you are clicking the link as is the google log output is cached by default in most browsers.
Therefore you can do one of the following: 
1. perform a hard refresh which for example on firefox would be ctrl+shift+r
2. If you have access to the project (ie in solo) then you can either take the ending part of the url and lookup the cloud build itself to get real time updates OR install the tamper monkey script that always redirects you. I like the manual approach as the tamper monkey scripts means you dont hit the outside of org pain which means build bot keeps living

davidjumani:
Thanks @nfuden. The tapermonkey script idea is quite helpful

puertomontt (approved):

PR #591, opened 2026-05-13
Bump proto gen timeout

sheidkamp:
Hitting this on an older M1 laptop. Check Point Endpoint Security may be playing a role in the slowness.


sheidkamp:
temporary issue, no fix needed

nfuden:
we always hit this XD :(

sheidkamp:
Oh, I thought I was the only one. I hit it immediately after a major version OS upgrade, and then it stopped happening.

We can reopen if helpful.

nfuden:
nah is good for now but I do see this happen periodically so having a sense when we are hitting it for real reasons versus an ephemeral reason is probably good eventually

PR #592, opened 2026-09-03
fix: keep xDS watches whose response has to be withheld

chandler-solo:
## Summary

Fix both paths that lose an xDS watch when ADS withholds a response because the client has not subscribed to every resource in the snapshot:

- `SetSnapshot` retains the unanswered watch.
- `CreateWatch` registers a watch when its immediate response is withheld.

Once a compatible snapshot arrives, the retained watch receives it.

The fixes ship **disabled by default**, controlled by `SOLOKIT_RETAIN_UNANSWERED_XDS_WATCHES`:

| Value | Behavior |
|---|---|
| `on` | Enable both fixes |
| `off` | Preserve previous behavior |
| `default`, empty, or unset | Follow the release default, currently disabled |
| Any other value | Log a warning and follow the release default |

The setting is read when each snapshot cache is created. The release default can change later; explicit `on` and `off` remain pinned.

Supersedes #586 by covering both the snapshot-update and request paths.

### DIFF
--- a/pkg/api/v1/control-plane/cache/simple.go
+++ b/pkg/api/v1/control-plane/cache/simple.go
@@ -168,9 +168,17 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) {
 				// empty resource - intended behavior is for the resources to be cleared
 				if resources != nil {
 					// snapshot has been initialized and exists
-					cache.respond(watch.Request, watch.Response, resources, version)
+					if !cache.respond(watch.Request, watch.Response, resources, version) {
+						// The response was withheld (see respond): keep the watch so a
+						// later snapshot that this request can accept is delivered to it.
+						if cache.log != nil {
+							cache.log.Debugf("keeping open watch priority %d and index %d :%v; version %q could not be sent",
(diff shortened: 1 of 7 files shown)


pkg/api/v1/control-plane/cache/simple.go
```diff
@@ -100,8 +126,38 @@ type CacheSettings struct {
 	PrioritySet map[int][]string
 }
 
+// RetainUnansweredXDSWatchesEnv controls retention of withheld ADS watches.
+// "on" and "off" pin the behavior; "default", unset, and empty follow the
+// release default. Invalid values warn and also follow the release default.
+const RetainUnansweredXDSWatchesEnv = "SOLOKIT_RETAIN_UNANSWERED_XDS_WATCHES"
```
andy-fong: maybe just call this ENV "EXP_FIX", I know it's very generic but if we are doing similar in other area to gate risky fix, we can re-use this. The cons is of course all or nothing. I don't think we need to keep it for long. The fix itself looks good to me.

chandler-solo: Done.

andy-fong (approved):

andy-fong (approved):

